Issue #30, opened 2017-02-24
isarray warns on install


Arcanemagus:
The `isarray` dependency throws a warning on install to just use `Array.isArray()` directly.

As I don't see an `engines` field in the `package.json` saying this needs to be compatible with an ancient version of Node.js that doesn't have that, it can probably just be removed.

wooorm:
Yeah I saw that. Pff, it sucks.
I’m trying to keep my stuff working without polyfills though, caring about old browsers too.
So for now I’m keeping it in!

PR #31, opened 2017-04-29
Remove deprecated isarray usage

just-boris:
This package is deprecated shows warning on `npm install`:

```
isarray@2.0.1: Just use Array.isArray directly
```

So I did that.


wooorm:
Thanks @just-boris, I removed it as it’s a bit frustrating, but I’d rather not use `isArray`: https://github.com/unifiedjs/unified/issues/30.

Issue #1, opened 2015-10-17
Component/duo support, try catch?


evansolomon:
Can you explain the idea behind the try/catch for a module that isn't part of the package in 3eacf6a75eaf247aa34a917906c9714d028d0871? I get that the result is that parent apps may supply their own copy of node-extend by depending on it themselves, but I can't tell why they'd want to do that. I also don't know what "component/duo support" means, and it's hard to work out what's going on just from the diff.

My use case is that I have this module in my dependency tree (via mdast) in a React Native app, and the React Native packager doesn't really love modules that cannot be resolved, so I'm trying to figure out the best way to handle this. It looks like the [node-extend](https://www.npmjs.com/package/node-extend) module is quite old and the github repo doesn't even exist anymore, so I don't particularly want to add it to my app.

Thanks.

wooorm:
Read more about [Component](https://github.com/componentjs/component) and [Duo](http://duojs.org). Read more about installation methods for mdast in its [documentation](https://github.com/wooorm/mdast/blob/master/doc/getting-started.md).

This work-around is needed because the `extend` npm module is called `node-extend` on GitHub (the project called `node-extend` on npm is never used). GitHub is used by non-npm package managers.

I have no experience with “React Native packager”. It doesn’t yet seem to support the `browser` field, but [they seem to be working on it](https://github.com/facebook/react-native/search?q=browser+field&type=Issues).

I’m closing this because I think supporting the browser field in **react-native/packager** is the cleanest fix for this.

An alternative is the [`blacklistRE`](https://github.com/facebook/react-native/tree/master/packager#reactpackagermiddlewareoptions) found in packager

evansolomon:
Thanks for the response. Unfortunately I don't think the blacklist option is accessible from the packager CLI. For now I fixed the issue by just adding unified v2.1.0 to the root project's dependencies.

PR #40, opened 2018-11-04
feature: support using unified as a template tag.

ChristianMurphy:
Inspired by [hyperHTML](https://github.com/WebReflection/hyperhtml) and [lit-html](https://github.com/polymer/lit-html).
This PR adds new `templateTag` and `templateTagSync` functions.

These new functions allow unified to be applied as a tag/modifier function to any template string.

Simple Example:

```js
var unified = require('unified')
var markdown = require('remark-parse')
var remark2rehype = require('remark-rehype')
var doc = require('rehype-document')
var format = require('rehype-format')
var stringify = require('rehype-stringify')

var html = unified()
  .use(markdown)
  .use(remark2rehype)
  .use(doc)
  .use(format)
  .use(stringify)
  .templateTagSync

console.log(html`# Hello world!`.toString())
```

Yields:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
  </head>
  <body>
    <h1>Hello world!</h1>
  </body>
</html>
```

Interpolation Example:

```js
var unified = require('unified')
var markdown = require('remark-parse')
var remark2rehype = require('remark-rehype')
var doc = require('rehype-document')
var format = require('rehype-format')
var stringify = require('rehype-stringify')

var html = unified()
  .use(markdown)
  .use(remark2rehype)
  .use(doc)
  .use(format)
  .use(stringify)
  .templateTagSync

var file = html`
# Hello world!
${['First', 'Second', 'Third'].map(header => `## ${header} subsection`).join('\n')}
`

console.log(file.toString())
```

Yields:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
  </head>
  <body>
    <h1>Hello world!</h1>
    <h2>First subsection</h2>
    <h2>Second subsection</h2>
    <h2>Third subsection</h2>
  </body>
</html>
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,6 +66,8 @@ function unified() {
   processor.runSync = runSync
   processor.process = process
   processor.processSync = processSync
+  processor.templateTag = templateTag
+  processor.templateTagSync = templateTagSync
 
   // Expose.
   return processor
@@ -396,6 +398,16 @@ function unified() {
       complete = true
       bail(err)
     }
+  }
+
+  // Fold template expression into a string that can be passed to `process`
+  function templateTag() {
+    return process(String.raw.apply(null, arguments))
+  }
+
+  // Fold template expression into a string that can be passed to `processSync`
+  function templateTagSync() {
+    return processSync(String.raw.apply(null, arguments))
   }
 }
 
--- a/readme.md
+++ b/readme.md
@@ -73,6 +73,8 @@ no issues found
     *   [processor.runSync(node\[, file\])](#processorrunsyncnode-file)
     *   [processor.process(file|value\[, done\])](#processorprocessfilevalue-done)
     *   [processor.processSync(file|value)](#processorprocesssyncfilevalue)
+    *   [processor.templateTag(strings\[, ...keys\])](#processortemplatetagstrings-keys)
+    *   [processor.templateTagSync(strings\[, ...keys\])](#processortemplatetagsyncstrings-keys)
     *   [processor.data(key\[, value\])](#processordatakey-value)
     *   [processor.freeze()](#processorfreeze)
 *   [Plugin](#plugin)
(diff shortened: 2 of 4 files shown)


wooorm:
Wow, this is super interesting!

I’m mainly wondering though, why do this? As in, your first example could be rewritten like so:

```js
var unified = require('unified')
var markdown = require('remark-parse')
var remark2rehype = require('remark-rehype')
var doc = require('rehype-document')
var format = require('rehype-format')
var html = require('rehype-stringify')

var html = unified()
  .use(markdown)
  .use(remark2rehype)
  .use(doc)
  .use(format)
  .use(html)
  .processSync

console.log(html(`# Hello world!`).toString())
```

Note the `processSync` and the parens. As they’re so similar, what’s the benefit of having this?

ChristianMurphy:
As it stands it's mostly syntax sugar to avoid needing `()`. :slightly_smiling_face: 

Where it could go is more interesting, similar to how MDX allows adding non-markdown content mid-document, supporting template literals opens up the possibility of being able to mixin new content types into documents without needing a transpiler.

wooorm:
@ChristianMurphy Right! So it gets interesting when it would be possible to “evaluate” the tag and the template in a build step?

ChristianMurphy:
My thought would be to use data type to determine how to evaluate.
`typeof string` would evaluate using current parser.
`typeof object | function`, would switch to a different processing track.

ChristianMurphy:
> `typeof object | function` would switch to a different processing track

In the alternative track, unified would parse the content surrounding the interpolated value, then add the interpolated value in as a second pass, directly into the AST.

E.G.

```js
const h = require('hastscript');

md2ht`## ${h('samp', 'sample header')}`
```

should generate

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
  </head>
  <body>
    <h2>
      <samp>
        sample header
      </samp>
    </h2>
  </body>
</html>

wooorm:
@ChristianMurphy Huhh, very interesting indeed! I like where this is heading. But I foresee that a) this feature would change in the future, as b) we’re not entirely sure how this would work yet.

It could of course be that first we need the implementation before we can create interesting cases, but I feel like it’s better to develop both together. Develop both the implementation and the use-cases together?

ChristianMurphy:
> I feel like it’s better to develop both together. Develop both the implementation and the use-cases together?

I agree.
Personally, I see this as the first step toward being able try out potential use cases.
If it should start as an external wrapper for unified, rather than a core processor function, that's fine.
This PR is mostly to get a discussion started.

wooorm:
@hypercubed interested in hearing your thoughts on this!

Hypercubed:
It's very interesting, and as you discussed, when you add the value interpolation.  In lit-html they have directives that can do may cool things like handling async functions.  At the moment I have couple of comments:

1) The unified transformations are text and static.  If the final result is text I don't really see directives being a help.  This is unlike lit-html where directives can do interesting things when rending the dom.  lit-html keep track of sstatic and dynamic parts of the template to allow effecient DOM updates.  Thsi simple has no benifit when generating static strings.

2) In lit-html the static and dynamic parts of the template are handeled very differenly.  The static parts must be complete html elements (for example html`<h1>${name}` + html`<h2>` doesn't work).  You cannot use dynamic content to generate elements (for example html`<${tag}>${name}<${tag}>`) for theis reason and also because the dynamic parts of HTML escaped.  The reason is in my point #1...  lit-html needs to seperate static and dynamic parts for effeicent updates.

3) In yoru example md2ht`## ${h('samp', 'sample header')}` you are mixing md and html... this can be an issue.  The function would need to know teh final product is html.

All this said I think it would be very intertsing in unified was able to convert to something like `lit-html`... then using a unified `md2lit` template method would be 💯

ChristianMurphy:
Thanks for the feedback @wooorm and @Hypercubed!

> something like lit-html... then using a unified `md2lit` template method would be :100:

It would be.
For that, it'd be nice to have incremental parsing, allowing more targeted updates of content.

PR #41, opened 2018-11-13
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
Current LTS branches are 6, 8 and 10. So we should test them all to make sure that nothing is breaking there (due to dependencies).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
-- lts/boron
+- 6
+- 8
+- 10
 - node
 after_script: bash <(curl -s https://codecov.io/bash)
 deploy:


wooorm:
6 in there by using `lts/boron`, and the latest node is in there as well. I’m not sure if it’s really needed to have everything in between there as well? (p.s. this one may be helpful:https://github.com/nodejs/Release#release-schedule)

DanielRuf:
`lts/boron` in general is not readable for most people and there are differences from 8 onwards.

Not all but some of them can have regressions or different outcomes and in this case we can catch the issues early and easily see if more are affected.

We have this (6 - 8 - 10 - node) in many projects.

wooorm:
> lts/boron in general is not readable for most people and there are differences from 8 onwards.

I’d personally disagree, I used to do that but switching made it *more* readable I think.

> Not all but some of them can have regressions or different outcomes and in this case we can catch the issues early and easily see if more are affected.

I guess, but honestly I’ve never had that happen. I used to have more versions in there, but over the last ±5 years, that never caught anything. What does happen is that Travis build sometimes randomly stall, so having more node versions == more builds == more stalls, and I think that’s a downside

wooorm:
@DanielRuf Thanks for the idea, but I believe we’re fine on the min/max versions, I think it’s the right trade-off between fast builds / fail-proof and catching all the bugs.

PR #43, opened 2018-12-11
Add type definitions

Rokt33r:
This is the first try to implement type definitions for unified.

I knew there is another attempt to adding types. But, I don't think the processor should persist AST type. It would make users confused because it looks tricky and doesn't work well with presets.

https://github.com/hrajchert/unified-typings-test/blob/master/src/types/unified/unified.d.ts

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "xo": "^0.23.0"
   },
   "scripts": {
-    "format": "remark . -qfo && prettier --write '**/*.js' && xo --fix",
+    "format": "remark . -qfo && prettier --write '**/{*.js,*.ts}' && xo --fix",
     "build-bundle": "browserify index.js -s unified > unified.js",
     "build-mangle": "esmangle unified.js > unified.min.js",
     "build": "npm run build-bundle && npm run build-mangle",
--- a/package.json
+++ b/package.json
@@ -25,6 +25,8 @@
     "lib"
   ],
   "dependencies": {
+    "@types/unist": "^2.0.0",
+    "@types/vfile": "^3.0.0",
     "bail": "^1.0.0",
     "extend": "^3.0.0",
     "is-plain-obj": "^1.1.0",
@@ -34,12 +36,14 @@
   },
   "devDependencies": {
     "browserify": "^16.0.0",
+    "dtslint": "^0.4.1",
     "esmangle": "^1.0.0",
     "nyc": "^13.0.0",
     "prettier": "^1.12.1",
     "remark-cli": "^6.0.0",
     "remark-preset-wooorm": "^4.0.0",
     "tape": "^4.4.0",
+    "typescript": "^3.2.2",
     "xo": "^0.23.0"
   },
   "scripts": {
@@ -49,7 +53,8 @@
     "build": "npm run build-bundle && npm run build-mangle",
     "test-api": "node test",
     "test-coverage": "nyc --reporter lcov tape test",
-    "test": "npm run format && npm run build && npm run test-coverage"
+    "test-types": "dtslint types",
+    "test": "npm run format && npm run build && npm run test-coverage && npm run test-types"
   },
   "nyc": {
     "check-coverage": true,
--- a/package.json
+++ b/package.json
@@ -20,7 +20,9 @@
   "contributors": [
     "Titus Wormer <tituswormer@gmail.com> (https://wooorm.com)"
   ],
+  "types": "types/index.d.ts",
   "files": [
+    "types/index.d.ts",
     "index.js",
     "lib"
   ],
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -120,6 +120,10 @@ declare namespace unified {
      * Get or set information in an in-memory key-value store accessible to all phases of the process.
      * An example is a list of HTML elements which are self-closing, which is needed when parsing, transforming, and compiling HTML.
      *
+     * @returns key-value store object
+     */
+    data(): {[key: string]: unknown}
+    /**
      * @param key Identifier
      * @returns If getting, the value at key
      */
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -25,7 +25,7 @@ declare namespace unified {
      */
     use(preset: Preset): Processor
     /**
-     * @param pluginTuple paris, plugin and options in an array
+     * @param pluginTuple pairs, plugin and options in an array
      */
     use(pluginTuple: PluginTuple): Processor
     /**
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -39,7 +39,7 @@ declare namespace unified {
      * @param file VFile or anything which can be given to vfile()
      * @returns Syntax tree representation of input.
      */
-    parse(file: vfile.VFile | string | Buffer): Unist.Node
+    parse(file: VFileCompatible): Unist.Node
 
     /**
      * Function handling the parsing of text to a syntax tree.
@@ -58,7 +58,7 @@ declare namespace unified {
      * @param file `VFile` or anything which can be given to `vfile()`
      * @returns String representation of the syntax tree file
      */
-    stringify(node: Unist.Node, file?: vfile.VFile): string
+    stringify(node: Unist.Node, file?: VFileCompatible): string
 
     /**
      * Function handling the compilation of syntax tree to a text.
@@ -81,9 +81,9 @@ declare namespace unified {
      * @returns `Promise` if `done` is not given. Rejected with an error, or resolved with the resulting syntax tree.
      */
     run(node: Unist.Node): Promise<Unist.Node>
-    run(node: Unist.Node, file: vfile.VFile): Promise<Unist.Node>
+    run(node: Unist.Node, file: VFileCompatible): Promise<Unist.Node>
     run(node: Unist.Node, done: RunCallback): void
-    run(node: Unist.Node, file: vfile.VFile, done: RunCallback): void
+    run(node: Unist.Node, file: VFileCompatible, done: RunCallback): void
 
     /**
      * Transform a syntax tree by applying plugins to it.
@@ -94,7 +94,7 @@ declare namespace unified {
      * @param file `VFile` or anything which can be given to `vfile()`
      * @returns The given syntax tree.
      */
-    runSync(node: Unist.Node, file?: vfile.VFile): Unist.Node
+    runSync(node: Unist.Node, file?: VFileCompatible): Unist.Node
 
     /**
      * Process the given representation of a file as configured on the processor. The process invokes `parse`, `run`, and `stringify` internally.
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -20,11 +20,15 @@
   "contributors": [
     "Titus Wormer <tituswormer@gmail.com> (https://wooorm.com)"
   ],
+  "types": "types/index.d.ts",
   "files": [
+    "types/index.d.ts",
```
wooorm: Any reason to keep everything in `types`? Do we except more types?

types/index.d.ts
```diff
@@ -0,0 +1,219 @@
...
+     * @param options Configuration for plugin]
+     * @returns The processor on which use is invoked
+     */
+    use(plugin: Plugin, options?: unknown): Processor
+    /**
+     * @param preset `Object` with an optional plugins (set to list), and/or an optional settings object
+     */
+    use(preset: Preset): Processor
+    /**
+     * @param pluginTuple paris, plugin and options in an array
```
wooorm: Paris? 🥖🧀🍷👨🏻‍🎨

```suggestion
     * @param pluginTuple pairs, plugin and options in an array
```

types/index.d.ts
```diff
@@ -0,0 +1,219 @@
...
+    processSync(file: vfile.VFile | string | Buffer): vfile.VFile
+
+    /**
+     * Get or set information in an in-memory key-value store accessible to all phases of the process.
+     * An example is a list of HTML elements which are self-closing, which is needed when parsing, transforming, and compiling HTML.
+     *
+     * @param key Identifier
+     * @returns If getting, the value at key
+     */
+    data(key: string): unknown
```
wooorm: [`data(): Object<string: unknown>`](https://github.com/unifiedjs/unified/blob/7dc311b1aa07d950549965c883e6fc1b1ed884a2/index.js#L133-L157) or something is also supported. But apparently not documented, so makes sense to ignore it for now. Or we need to document it as well.

types/index.d.ts
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  type Plugin = Attacher
+  /**
+   * Presets provide a potentially sharable way to configure processors.
+   * They can contain multiple plugins and optionally settings as well.
+   */
+  interface Preset {
+    plugins?: PluggableList
+    settings?: unknown
```
wooorm: `settings` must be an object when set

types/unified-tests.ts
```diff
@@ -0,0 +1,164 @@
+import * as Unist from 'unist'
```
wooorm: Does it make sense to move these to `test/`?

types/index.d.ts
Rokt33r: I'll update document too. 👍

types/index.d.ts
Rokt33r: I'll fix it.

types/index.d.ts
Rokt33r: 👍

Rokt33r:
> In all cases where a human can give a vfile, it could also be a Buffer or string. You added that in most places I think. But, maybe we should use VFileOptions there. As it supports some other stuff as well. Does it make sense to export VFileOptions from vfile, and use it here instead of this list of VFile | Buffer | string?
Hmm... To be honestly, I don't like to give lots of options in api. It usually produce more chaos in my experience. I actually don't like `processor.data()` too because it smells like jQuery. Its behavior is changed by the number of arguments. But, it is just my opinion. I don't want to push you. So, let me know if you want to change it. Then, I'll make another pr to fix it.

> Can we run prettier on the typescript files as well?
We can do. It works well with the current configuration. 👍
I'll update prettier script so it can format index.d.ts too.

types/index.d.ts
```diff
@@ -0,0 +1,228 @@
...
+
+  type RunCallback = (
+    error: Error | null,
+    node: Unist.Node,
+    file: vfile.VFile
+  ) => void
+
+  type ProcessCallback = (error: Error | null, file: vfile.VFile) => void
+
+  type VFileCompatible = vfile.VFile | vfile.VFileOptions | string | Buffer
```
wooorm:
```suggestion
  type VFileCompatible = vfile.VFile | vfile.VFileOptions | vfile.VFileContents
```

I think this works as well?

types/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "lib": ["es2015", "dom"],
```
wooorm: Do we use the DOM btw? Never wrote a tsconfig so not sure if it’s needed? 🤷‍♂️

types/index.d.ts
Rokt33r: LGTM! I'll replace it!

types/tsconfig.json
Rokt33r: We don't need it. And it doesn't matter for users because `tsconfig.json` is used for the type testing only. I'll delete it too.

types/index.d.ts
Rokt33r: Btw, I think we should define `VFileCompatible` to `vfile` later. Let's do it later when we move its types from DefinitelyTyped repo.

types/index.d.ts
```diff
@@ -0,0 +1,228 @@
+// TypeScript Version: 3.0
+
+import * as Unist from 'unist'
+import vfile = require('vfile')
+import {parse} from 'url'
+import {stringify} from 'querystring'
```
wooorm: Hey are we using these btw? 🤔

wooorm: Sorry, I missed them before 😬

Rokt33r: lol... I deleted...!!

wooorm:
Alright, so we should release this. As a minor I think? It’s not breaking in any way right?

Rokt33r:
It won't break anything. Release as minor version should be fine. 👍

wooorm:
Released! Thanks @Rokt33r, and also @hrajchert!

rjimenezda:
This broke a project I have that uses both stylelint and Stencil 0.13. The dependency chain made me end up here.

So any projects that were using Typescript 2.x could be broken? I might have something wrong on the compiler settings, though. As I said, this gets installed on my package because of stylelint, so I don't know why tsc would parse these definitions :thinking:

wooorm:
Huh, weird, @Rokt33r, any thought on this?

ChristianMurphy:
@rjimenezda 

> This broke a project I have

What is the error message/stacktrace?

> I might have something wrong on the compiler settings, though.

What is the `tsconfig` for the project?

ChristianMurphy:
related to https://github.com/vfile/vfile-message/issues/3

ChristianMurphy:
Discussion on supported TypeScript versions started on Spectrum https://spectrum.chat/unified/type-definitions?thread=7c1f6266-6be2-4d0a-b599-4ff93a6d0efd

rjimenezda:
@ChristianMurphy 

It's complicated, because Stencil uses typescript under the hood, so I'm not 100% sure of the settings. If you're curious, it's [here](https://github.com/CartoDB/airship/tree/master/packages/components)

Just to verify, I set up a quick project with typescript 2.9 and stylelint as dependencies with a simple tsconfig:

```json
{
    "compilerOptions": {
        "module": "system",
        "noImplicitAny": true,
        "removeComments": true,
        "preserveConstEnums": true,
        "outFile": "out.js",
        "sourceMap": true
    },
    "include": [
        "src/**/*"
    ],
    "exclude": [
        "node_modules",
        "**/*.spec.ts"
    ]
}
```

If you run tsc with this config the following error a few times:

```
> tsc
node_modules/@types/unist/index.d.ts:31:20 - error TS2304: Cannot find name 'unknown'.

31     [key: string]: unknown;
```

On the sample project, the build actually finishes. Stencil seems to stop what it's doing if it encounters an error, though.

In any case, I **was able to suppress this error** by adding `"skipLibCheck": true` to the compiler options. This is kind of annoying (TS's fault) because I can't be selective about it.

It seems like tsc somehow automatically parses all definitions in node_modules/@types, because I have at least one other .d.ts files in node_modules using 'unknown' and it's being ignored.

I've actually checked the DefinitelyTyped repo and there's only a handful of projects that use `unknown`, so I wonder if this will gradually become more of common issue. I can't find anything regarding them having a way to control typescript versions.

The worst part about this is that I'm not even actually using this library, otherwise I'd be super happy to update and get the definition files 😂

ChristianMurphy:
> It's complicated, because Stencil uses typescript under the hood, so I'm not 100% sure of the settings.

This should be resolved by https://github.com/ionic-team/stencil/commit/127f0d574eaace1def5e4f3cf1e2088cff37266a in stencil version 0.16.1 and above.

rjimenezda:
Of course it will. We haven't upgraded yet, but we're planning on doing so eventually.

This will still happen to every single project that uses typescript < 3 and for whatever reason, this library is in their dependency graph.

In any case, I'll keep this in mind the next time I have to add types to a library through @types and make it a major, to minimize cases like mine.

Rokt33r:
@wooorm It seems to be possible to move `index.d.ts` to the root dir of the project.

Like the below:

`test/types/test.ts`
`test/types/tsconfig.json`
`test/types/tslint.json`
`index.d.ts`

by updating `tsconfig.json`:

```
{
  "compilerOptions": {
    "lib": ["es2015"],
    "strict": true,
    "noImplicitAny": true,
    "noImplicitThis": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "baseUrl": ".",
    "paths": {
      "unified": ["../../index.d.ts"] // <-- here
    }
  }
}
```

If the file structure looks good, I'll fix this too after we resolve the versioning problem.

Ore4444:
Yes, adding native typings can break projects that coded their own typings for Unified

For example, it broke my Unified service which had the following:

```ts
type UnifiedPluginFn = (opts: unknown) => (tree: any) => void
type UnifiedPlugin = UnifiedPluginFn | [UnifiedPluginFn, any]

interface IUnifiedProcessor {
    use: (plugin: UnifiedPlugin, opts?: unknown) => void
    process: (htmlInput: string) => any // vfile
}
```

Of course, `IUnifiedProcessor` is not compatible with `unified.Processor`.
nvm i fixed it.

artberri:
Hello,

As it is said in https://github.com/unifiedjs/unified/pull/43#issuecomment-447500058 and https://github.com/unifiedjs/unified/pull/43#issuecomment-447504706 this upgrade will affect "every single project that uses typescript < 3 and for whatever reason, this library is in their dependency graph". 

This happened to a project I'm working on because it depends on `stylelint`, and now we need to `freeze` the unified version because we don't want to add `"skipLibCheck": true` and we can't upgrade to Typescript 3 yet (I'm investigating if by using the `types` option in the compiler options I can be more selective than using skipLibCheck). 

I see some discussions, like this https://github.com/Microsoft/TypeScript/issues/9731, about if @types should be a dependency, a dev dependency or a peer dependency in order to avoid these issues but I'm not sure if there is a "good" answer yet.

Are you thinking on doing something on this project (like bumb new mayor version, or remove the types as direct dependencies in 7.1.1,...)? Any suggestion on what should we do?

Thank you in advance.

SwinX:
I'm sorry for not just upvoting comment above, but this problem is really annoying. We encountered exactly same issue because we're distributing so-called test-utilities project internally among several microservices, so we need to have `stylelint`, which is using this package explicitly, as direct package dependency.
It would be really great if you will alter typings and add support of typescript < 3.0.

ChristianMurphy:
See the TypeScript definition discussion here https://spectrum.chat/?t=7c1f6266-6be2-4d0a-b599-4ff93a6d0efd

PR #44, opened 2019-01-22
Remove use of unknown from typings

SwinX:
This is needed to be able to depend on unified in projects which are not supporting typescript 3.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -17,7 +17,7 @@ declare namespace unified {
      * @param options Configuration for plugin]
      * @returns The processor on which use is invoked
      */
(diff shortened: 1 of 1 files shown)


ChristianMurphy:
@SwinX See https://spectrum.chat/?t=7c1f6266-6be2-4d0a-b599-4ff93a6d0efd

ChristianMurphy:
Thanks @SwinX!
Currently Unified plans to support TypeScript 3 or newer typings.
For TypeScript 2 or Typescript 1 support Unified version 7.0.2 can be pinned in dependencies.

PR #46, opened 2019-04-17
Add Spectrum info on how to get help

diessica:
The introductory README lacks information on how to get help when using Unified. This pull requests  orientate users on how to get their questions answered or help others by pointing them to our already existing **[Unified community on Spectrum](https://spectrum.chat/unified)**.

The community itself is already in the README but I wasn't aware it was a thing – it might be the case for others too. Hopefully this will increase visibility!

PR #49, opened 2019-06-09
refactor typings

ChristianMurphy:
Refactor typings, drop `declare` statements causing #46 to fail CI.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,226 +1,218 @@
 // TypeScript Version: 3.0
 
-import * as Unist from 'unist'
-import vfile = require('vfile')
+import {Node} from 'unist'
+import vfile, {VFile, VFileOptions, VFileContents} from 'vfile'
 
-declare namespace unified {
-  interface Processor {
-    /**
-     * @returns New unfrozen processor which is configured to function the same as its ancestor. But when the descendant processor is configured in the future it does not affect the ancestral processor.
-     */
-    (): Processor
-
-    /**
-     * Configure the processor to use a plugin and optionally configure that plugin with options.
-     *
-     * @param plugin unified plugin
-     * @param options Configuration for plugin]
-     * @returns The processor on which use is invoked
-     */
-    use(plugin: Plugin, options?: unknown): Processor
-    /**
-     * @param preset `Object` with an optional plugins (set to list), and/or an optional settings object
-     */
-    use(preset: Preset): Processor
-    /**
-     * @param pluginTuple pairs, plugin and options in an array
-     */
-    use(pluginTuple: PluginTuple): Processor
-    /**
-     * @param list List of plugins, presets, and pairs
-     */
-    use(list: PluggableList): Processor
-
-    /**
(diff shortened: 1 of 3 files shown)


types/index.d.ts
```diff
@@ -1,226 +1,218 @@
...
-    (this: Processor, options?: unknown): Transformer | void
-  }
-
-  /**
-   * Transformers modify the syntax tree or metadata of a file. A transformer is a function which is invoked each time a file is passed through the transform phase.
-   * If an error occurs (either because it’s thrown, returned, rejected, or passed to `next`), the process stops.
-   *
-   * The transformation process in unified is handled by `trough`, see it’s documentation for the exact semantics of transformers.
+import {Node} from 'unist'
+import vfile, {VFile, VFileOptions, VFileContents} from 'vfile'
```
Rokt33r: All of the modules in unified ecosystem are using cjs. We should import it like `import something = require('something')`. https://github.com/Microsoft/dtslint/blob/master/docs/no-import-default-of-export-equals.md

So we have two choices.

1. Retrieve types like `vfile.*`.

```ts
import vfile = require('vfile')

// Now we can use the interfaces like `vfile.VFile`
```

2. Have 2 import statements.

```ts
import vfile = require('vfile')
import {VFile, VFileOptions, VFileContents} from 'vfile'
```

ChristianMurphy: @Rokt33r theres the conundrum.
the
```ts
import vfile = require('vfile')
```

style imports are part of what is causing CI on master branch to fail.
See the logs in #46.

wooorm: I found that the linter is overly critical. We could turn off some of the linting maybe?

Rokt33r: @Rokt33r 

It works fine for me. I submitted a pr to this pr. https://github.com/ChristianMurphy/unified/pull/1

It might be dependency problems.

```
unified@7.1.0 /Users/junyoung/Code/unified
├─┬ dtslint@0.4.9
│ └── typescript@3.6.0-dev.20190608 
└── typescript@3.5.1
```

If you are still getting an error, could you share the error log?

Rokt33r: > I found that the linter is overly critical. We could turn off some of the linting maybe?

@wooorm We can do by modifying `types/tslint.json`.

ChristianMurphy: merged, thanks @Rokt33r!

wooorm:
@Rokt33r Do you approve this now?

Rokt33r:
@wooorm LGTM!

JamesMessinger:
These typing changes don't work due to the incorrect use of `export default` instead of `export =`.

Unified **doesn't** export an ESM module.  If it did, then these type definitions would be valid.  But Unified exports a CommonJS module, so the correct type definitions would be:

```typescript
export = function unified(): Processor
```

See [the TypeScript documentation](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) for more details

Issue #45, opened 2019-03-07
@types packages should be in devDependencies


damonmaria:
### Subject of the issue

This package includes `@types/unist` and `@types/vfile` as *dependencies*

### Your environment

*   MacOS 10.14.3
*   node v10.15.2, yarn 1.13.0
*   Package `react-scripts@2.1.6` (from Create React App) includes unified

### Steps to reproduce

1. `yarn add unified`
2. `yarn why @types/node`

### Expected behaviour

No `@types` packages should be installed from adding unified. 

### Actual behaviour

Adding unified adds `@types/unist`, `@types/vfile`, `@types/node` and other packages into node_modules.

Rokt33r:
@damonmaria @ChristianMurphy We've discussed this issue in here.

https://spectrum.chat/unified/type-definitions/where-should-we-put-types-unist~a22fb3db-88bf-472e-8cbb-9e4191116d6d

I also agree with @damonmaria 's opinion.

So I think the best solution is:
- Move `@types/unist` to `devDependencies` (without adding it `peerDependencies`)
- Add description, which is installing `@types/unist`, to readme.md or other place for typescript users.

damonmaria:
The actual base issue I'm having is that @types/node, @types/react-native and @types/styled-components in some combination in different versions declare 'base things' (like `console` and `FormData`) differently. Which causes Typescript to bork. I'm in a web app so `@types/node` shouldn't even be there. But at the moment `react-scripts` is forcing that to happen (through this package). 

Sorry, I can't speak to what `unified` should do as I'm not even a direct user of it.

Rokt33r:
`unified` is a pluggable interface. It accept a parser, a compiler and some transformers. For example, `remark` is actually a unified instance with `remark-parse` and `remark-compiler`.

I think dedupe might help. https://yarnpkg.com/lang/en/docs/cli/dedupe/

But, it seems you need to install @types/node when using cra with typescript. https://facebook.github.io/create-react-app/docs/adding-typescript

damonmaria:
Yes, I've seen the CRA recommendation for installing `@types/node` but it seems crazy since in a CRA app you *don't* have all of Node available. With `@types/node` hanging about you can do things like `import { readFile } from 'fs'` which in a web app, well... obviously isn't going to work.

I always switch `@types/node` out for `@types/webpack-env` which declares `process.env` and the webpack internals. `process.env` is all I think from `@types/node` which is actually available in a CRA/Webpack app.

[...]

PR #52, opened 2019-06-14
Move types from dependencies to dev-dependencies

wooorm:
Closes GH-45.

/cc @Rokt33r


ChristianMurphy:
Fine by me

> It's worth noting that this issue does not appear to happen when using Create React App version 3.

https://github.com/unifiedjs/unified/issues/45#issuecomment-502163253

Issue #47, opened 2019-05-14
Return type of `stringify` could be not a `string`


jaens:
The TypeScript definition[ specifies it as string](https://github.com/unifiedjs/unified/blob/0dc92d6397a7fbbe70a93475db1bd585e12d1845/types/index.d.ts#L59).
But - if you use eg. `rehype-react` as the last stage of the pipeline, then the return type is actually a `React.VNode`.
This causes TS to issue an error.

Not sure what's the best solution here.
* Specify the "stringed" type as a generic parameter somewhere?
* Make the return type `any`?

ChristianMurphy:
@jaens could you give an example of when this happens?
The docs in `rehype-react` show `process`and `processSync` https://github.com/rhysd/rehype-react#use
Using the `process` methods may work better with TS typings.

---

I'm cautious about using `stringify` to return a non-string or setting it's typescript type to non-string, as this seems to run counter to its intended use.

wooorm:
`.stringify` is indeed “supposed” to return a string. Or at least the “compiled” syntax tree. All the compilers that go to a virtual DOM (remark-vdom, remark-react, rehype-react, etc) benefit from the fact that it doesn’t have to be.

In fact, anything returned by `compiler` (or `Compiler#compile`) is returned by `.stringify`.
When using `.process`, that value is set on `file.contents`, which typically would be `string` or `Buffer`, but can also be anything.

---

> I'm cautious about using stringify to return a non-string or setting it's typescript type to non-string, as this seems to run counter to its intended use.
> — [ChristianMurphy](https://github.com/unifiedjs/unified/issues/47#issuecomment-492072009)

Yeah, I feel the same way.

jaens:
This is for a two-stage pipeline, so `process` is not an option, I think.

The first stage runs the parser, extracts the metadata etc., and the second stage renders the result from the first stage into React (this is done in two stages because this is a multi-page renderer with interdependencies between pages).

The code is roughly:
```javascript
// Stage 1 (vfile as input, node as output)
let processor = unified()
      .use(remark_parse)
      .use(remark_frontmatter, "yaml")
      .use(frontMatterExtractor);

let node = await processor.run(await processor.parse(vfile), vfile);
```
```javascript
// Stage 2 (node as input, vfile as input, React as output)
let processor = unified().use(rehype_react, {
      createElement: React.h
});
return ((await processor.stringify(
      node,
      vfile
    )) as any) as React.VNode;
```

[...]

PR #53, opened 2019-06-14
Add support for non-string compiler results

wooorm:
Note: there are reservations for this approach: https://github.com/unifiedjs/unified/issues/47#issuecomment-500255507.

Closes GH-47.

/cc @Rokt33r

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -54,9 +54,9 @@ export interface Processor {
    *
    * @param node
    * @param file `VFile` or anything which can be given to `vfile()`
-   * @returns String representation of the syntax tree file
+   * @returns Typically a string representation of the syntax tree file, sometimes another syntax tree (such as a React node).
    */
-  stringify(node: Node, file?: VFileCompatible): string
+  stringify<P = string>(node: Node, file?: VFileCompatible): P
 
   /**
    * Function handling the compilation of syntax tree to a text.


wooorm:
Failure in dtslint is due to: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md

wooorm:
I don’t have an informed enough opinion on this matter or TypeScript, so I’d abstain from voting for or against it: I created this PR based on a suggestion by @Rokt33r to continue the conversation.

I’m interested in hearing about the downsides from Junyoung.

@ChristianMurphy Is casting a solution we can apply here in unified, or does that happen on the user’s side?

ChristianMurphy:
> Is casting a solution we can apply here in unified, or does that happen on the user’s side?

It would happen on the user's side, it's something we could document.

wooorm:
That doesn’t sound too bad...

wooorm:
@ChristianMurphy How should we document that? Here? https://github.com/unifiedjs/unified#returns-3

…maybe in also here? https://github.com/unifiedjs/unified#returns-6

…or should it go in *-react and the like other stringifier plugins?

ChristianMurphy:
> or should it go in *-react and the other stringifier plugins?

This option sounds good to me.
The custom type will be custom per plugin.

Rokt33r:
> or should it go in *-react and the other stringifier plugins?

Should be ideal.

Until it is released, we could notice users to do like `(processor.stringify(node, vfile) as unknown) as VNode`. as @ChristianMurphy suggested in https://github.com/unifiedjs/unified/issues/47#issuecomment-500255507 rather than merging this pr.

Rokt33r:
@wooorm And I'm afraid that my decision is right because I'm not using remark with typescript now. My work is stale because I don't know which version of typescript should I support. https://github.com/remarkjs/remark/pull/383#issuecomment-500189439

If you give some kind of confirmation, I could start working again. In my opinion, we can ignore old versions of typescript until the type definitions for unified ecosystem are settled. [spectrum: TypeScript version(s) supported?
](https://spectrum.chat/unified/type-definitions/typescript-version-s-supported~7c1f6266-6be2-4d0a-b599-4ff93a6d0efd)

wooorm:
Alright, I’ll close this PR.
We need a new PR closing GH-47 then, to add `(processor.stringify(node, vfile) as unknown) as VNode` to the docs here.

I’d appreciate if other people could do the same when needed to the other plugins returning non-strings.

> My work is stale because I don't know which version of typescript should I support. remarkjs/remark#383 (comment)

Sorry about that. I don’t have an informed opinion. Going with current (3?) seems good to me? Thanks for working on types!

Rokt33r:
I guess we need higher than v3.1 at least. Then, I'll ignore lower version support until types are settled to unified ecosystem! I could start working again from this weekend!

wooorm:
Awesome, sounds good to me!

PR #57, opened 2019-06-29
Corrected type definitions. Fixes #56

### LINKED ISSUE #56: Broken TypeScript definitions
### Subject of the issue

The TypeScript type definitions were refactored in PR #49, but the new definitions are invalid due to the use of `export default` instead of `export =`.  See [the TypeScript documentation](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) to understand the difference between the two.

### Your environment

*   **OS**: Windows
*   **Packages**: 
    - unified v8.0.0
    - typescript: 3.5.2
*   **Env**: 
    - Node v12.2.0
    - NPM v6.9.0

### Steps to reproduce

You can see this bug in the wild in [the rehype-inline-svg plugin](https://github.com/JS-DevTools/rehype-inline-svg).  

1. __Clone the repo__<br>
`git clone https://github.com/JS-DevTools/rehype-inline-svg.git`

2. __Install dependencies__<br>
`npm install`

3. __Build the code__<br>
`npm run build` <br>
The build will succeed because the repo is currently using Unified v7.1.0 which has the old TypeScript definitions, which correctly used `export =` syntax.

4. __Upgrade Unified__
`npm install unified@8`

5. __Build the code again__<br>
`npm run build`<br>
This time the build will fail because of the new type definitions with incorrect `export default` syntax.

### Expected behaviour

**Option 1**
The TypeScript definitions should correctly reflect the fact that Unified exports a CommonJS module, _not_ an ESM module.

**Option 2**
Change Unified to export an ESM module instead of, or in addition to a CommonJS module. This way, the new TypeScript definitions will be correct.

### Actual behaviour

TypeScript builds fail because the type definitions are incorrect.

JamesMessinger:
This PR fixes #56 by correcting the TypeScript definitions to properly reflect the fact that Unified exports a CommonJS module rather than an ESM module. 

* Replaced incorrect `export default`  syntax with `export =`
* Wrapped all the TypeScript type definitions in a namespace

### DIFF
--- a/types/tsconfig.json
+++ b/types/tsconfig.json
@@ -6,7 +6,6 @@
     "noImplicitThis": true,
     "strictNullChecks": true,
     "strictFunctionTypes": true,
-    "esModuleInterop": true,
     "baseUrl": ".",
     "paths": {
       "unified": ["index.d.ts"]
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,219 +1,222 @@
 // TypeScript Version: 3.0
 
-import {Node} from 'unist'
+import { Node } from 'unist';
+import { VFile, VFileContents, VFileOptions } from 'vfile';
 import vfile = require('vfile')
-import {VFile, VFileOptions, VFileContents} from 'vfile'
 
-export interface Processor {
+declare namespace unified {
+  interface Processor {
+    /**
+     * @returns New unfrozen processor which is configured to function the same as its ancestor. But when the descendant processor is configured in the future it does not affect the ancestral processor.
+     */
+    (): Processor
+
+    /**
+     * Configure the processor to use a plugin and optionally configure that plugin with options.
+     *
+     * @param plugin unified plugin
+     * @param options Configuration for plugin]
+     * @returns The processor on which use is invoked
+     */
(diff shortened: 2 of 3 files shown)


types/tsconfig.json
```diff
@@ -6,7 +6,6 @@
     "noImplicitThis": true,
     "strictNullChecks": true,
     "strictFunctionTypes": true,
-    "esModuleInterop": true,
```
ChristianMurphy: Would documenting that `esModuleInterop` should be enabled, resolve this issue without reverting the typings refactor?

JamesMessinger: The `esModuleInterop` setting should only be used in end applications, not libraries. It's a way for app developers to treat CommonJS modules as though they were ESM modules, but it results in lots of little helper functions being injected into the code to handle the mapping between the two module formats. App developers should be able to opt-in or opt-out of that behavior, depending on whether they're okay with the trade-off.

wooorm:
How should this be released?

Rokt33r:
@wooorm Did you mean the version? If so, we could publish as a patch version, like `8.0.1`.

wooorm:
Yeah, I meant version. So a patch is fine? Alright!

PR #58, opened 2019-07-02
allow plugins to declare an interface for their settings

ChristianMurphy:
An initial implementation for https://github.com/remarkjs/remark/pull/383#discussion_r299623495

> Unified types would highly benefit from a generic here. As written, options passed along with the use() call is a place where all type safety is lost.

This makes plugins generic, defaulting to unified's `Settings` type.
Plugins that do not specify a settings interface should *not* be unaffected.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -5,7 +5,7 @@ import {VFile, VFileContents, VFileOptions} from 'vfile'
 import vfile = require('vfile')
 
 declare namespace unified {
-  interface Processor {
+  interface Processor<T = Settings> {
     /**
      * @returns New unfrozen processor which is configured to function the same as its ancestor. But when the descendant processor is configured in the future it does not affect the ancestral processor.
      */
@@ -18,7 +18,7 @@ declare namespace unified {
      * @param options Configuration for plugin]
      * @returns The processor on which use is invoked
      */
-    use(plugin: Plugin, options?: unknown): Processor
+    use<T = Settings>(plugin: Plugin<T>, options?: T): Processor
     /**
      * @param preset `Object` with an optional plugins (set to list), and/or an optional settings object
      */
@@ -26,11 +26,11 @@ declare namespace unified {
     /**
      * @param pluginTuple pairs, plugin and options in an array
      */
-    use(pluginTuple: PluginTuple): Processor
+    use<T = Settings>(pluginTuple: PluginTuple<T>): Processor
     /**
      * @param list List of plugins, presets, and pairs
      */
-    use(list: PluggableList): Processor
+    use<T = Settings>(list: PluggableList<T>): Processor
 
     /**
      * Parse text to a syntax tree.
@@ -145,7 +145,7 @@ declare namespace unified {
     freeze(): Processor
   }
 
-  type Plugin = Attacher
+  type Plugin<T = Settings> = Attacher<T>
   type Settings = {
     [key: string]: unknown
   }
@@ -157,9 +157,9 @@ declare namespace unified {
     plugins?: PluggableList
     settings?: Settings
   }
-  type PluginTuple = [Plugin, Settings]
-  type Pluggable = Plugin | Preset | PluginTuple
-  type PluggableList = Pluggable[]
+  type PluginTuple<T = Settings> = [Plugin<T>, T]
+  type Pluggable<T = Settings> = Plugin<T> | Preset | PluginTuple<T>
+  type PluggableList<T = Settings> = Array<Pluggable<T>>
 
   /**
    * An attacher is the thing passed to `use`.
@@ -171,8 +171,8 @@ declare namespace unified {
    * @param options Configuration
    * @returns Optional.
    */
-  interface Attacher {
-    (this: Processor, options?: unknown): Transformer | void
+  interface Attacher<T = Settings> {
+    (this: Processor, options?: T): Transformer | void
   }
 
   /**
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -30,7 +30,7 @@ declare namespace unified {
     /**
      * @param list List of plugins, presets, and pairs
      */
-    use<T = Settings>(list: PluggableList<T>): Processor
+    use(list: PluggableList): Processor
 
     /**
      * Parse text to a syntax tree.
@@ -159,7 +159,7 @@ declare namespace unified {
   }
   type PluginTuple<T = Settings> = [Plugin<T>, T]
   type Pluggable<T = Settings> = Plugin<T> | Preset | PluginTuple<T>
-  type PluggableList<T = Settings> = Array<Pluggable<T>>
+  type PluggableList = Array<Pluggable<any>>
 
   /**
    * An attacher is the thing passed to `use`.
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -5,7 +5,7 @@ import {VFile, VFileContents, VFileOptions} from 'vfile'
 import vfile = require('vfile')
 
 declare namespace unified {
-  interface Processor<T = Settings> {
+  interface Processor {
     /**
      * @returns New unfrozen processor which is configured to function the same as its ancestor. But when the descendant processor is configured in the future it does not affect the ancestral processor.
      */
@@ -15,18 +15,33 @@ declare namespace unified {
      * Configure the processor to use a plugin and optionally configure that plugin with options.
      *
      * @param plugin unified plugin
-     * @param options Configuration for plugin]
+     * @param options Configuration for plugin
+     * @param extraOptions Additional configuration for plugin
      * @returns The processor on which use is invoked
      */
-    use<T = Settings>(plugin: Plugin<T>, options?: T): Processor
+    use<T = Settings, S = undefined>(
+      plugin: Plugin<T, S>,
+      options?: T,
+      extraOptions?: S
+    ): Processor
+
     /**
      * @param preset `Object` with an optional plugins (set to list), and/or an optional settings object
      */
     use(preset: Preset): Processor
+
     /**
      * @param pluginTuple pairs, plugin and options in an array
      */
     use<T = Settings>(pluginTuple: PluginTuple<T>): Processor
+
+    /**
+     * @param pluginTriple plugin, options, and extraOptions in an array
+     */
+    use<T = Settings, S = undefined>(
+      pluginTriple: PluginTriple<T, S>
+    ): Processor
+
     /**
      * @param list List of plugins, presets, and pairs
      */
@@ -145,7 +160,7 @@ declare namespace unified {
     freeze(): Processor
   }
 
-  type Plugin<T = Settings> = Attacher<T>
+  type Plugin<T = Settings, S = undefined> = Attacher<T, S>
   type Settings = {
     [key: string]: unknown
   }
@@ -158,8 +173,13 @@ declare namespace unified {
     settings?: Settings
   }
   type PluginTuple<T = Settings> = [Plugin<T>, T]
-  type Pluggable<T = Settings> = Plugin<T> | Preset | PluginTuple<T>
-  type PluggableList = Array<Pluggable<any>>
+  type PluginTriple<T = Settings, S = undefined> = [Plugin<T, S>, T, S]
+  type Pluggable<T = Settings, S = undefined> =
+    | Plugin<T>
+    | Preset
+    | PluginTuple<T>
+    | PluginTriple<T, S>
+  type PluggableList = Array<Pluggable<any, any>>
 
   /**
    * An attacher is the thing passed to `use`.
@@ -169,10 +189,11 @@ declare namespace unified {
    *
    * @this Processor context object is set to the invoked on processor.
    * @param options Configuration
+   * @param extraOptions Secondary configuration
    * @returns Optional.
    */
-  interface Attacher<T = Settings> {
-    (this: Processor, options?: T): Transformer | void
+  interface Attacher<T = Settings, S = undefined> {
+    (this: Processor, options?: T, extraOptions?: S): Transformer | void
   }
 
   /**
--- a/types/tslint.json
+++ b/types/tslint.json
@@ -9,6 +9,7 @@
     "semicolon": false,
     "unified-signatures": false,
     "whitespace": false,
-    "interface-over-type-literal": false
+    "interface-over-type-literal": false,
+    "no-unnecessary-generics": false
   }
 }
--- a/types/unified-tests.ts
+++ b/types/unified-tests.ts
@@ -33,7 +33,12 @@ interface ExamplePluginSettings {
 const typedPlugin: Plugin<ExamplePluginSettings> = function() {}
 const typedSetting = {example: 'example'}
 
-const implicitlyTypedPlugin = (settings?: ExamplePluginSettings) => {}  
+const implicitlyTypedPlugin = (settings?: ExamplePluginSettings) => {}
+
+const pluginWithTwoSettings = (
+  processor?: Processor,
+  settings?: ExamplePluginSettings
+) => {}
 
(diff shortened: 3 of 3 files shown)


types/index.d.ts
```diff
@@ -157,9 +157,9 @@ declare namespace unified {
     plugins?: PluggableList
     settings?: Settings
   }
-  type PluginTuple = [Plugin, Settings]
-  type Pluggable = Plugin | Preset | PluginTuple
-  type PluggableList = Pluggable[]
+  type PluginTuple<T = Settings> = [Plugin<T>, T]
+  type Pluggable<T = Settings> = Plugin<T> | Preset | PluginTuple<T>
+  type PluggableList<T = Settings> = Array<Pluggable<T>>
```
ChristianMurphy: There is an edge case here.
If two plugins declare different interfaces there will be a clash.

For example:

```ts
interface ExamplePluginSettings {
  example: string
}
interface ExamplePluginSettings2 {
  example2: string
}

const typedPlugin: Plugin<ExamplePluginSettings> = function() {}
const typedPlugin2: Plugin<ExamplePluginSettings2> = function() {}

const typedSetting = {example: 'example'}
const typedSetting2 = {example2: 'example'}

processor.use([[typedPlugin, typedSetting], [typedPlugin2, typedSetting2]])
```

will result in

```
Type '(Attacher<ExamplePluginSettings2> | { example2: string; })[]' is not assignable to type 'Pluggable<{ example: string; }>'.
  Type '(Attacher<ExamplePluginSettings2> | { example2: string; })[]' is missing the following properties from type '[Attacher<{ example: string; }>, { example: string; }]': 0, 1ts(2322)
```

There are a few options here:

1. set the generic to `unknown` or `any` and lose the strong typing in this case
2. leave the typing as is, typescripters can work around the issue by chaining `use` methods, one per typed plugin
3. something else?

types/tslint.json
```diff
@@ -9,6 +9,7 @@
     "semicolon": false,
     "unified-signatures": false,
     "whitespace": false,
-    "interface-over-type-literal": false
+    "interface-over-type-literal": false,
+    "no-unnecessary-generics": false
```
ChristianMurphy: Option 3 from https://github.com/unifiedjs/unified/pull/58/files#r299705792 would allow this rule to be re-enabled.

types/unified-tests.ts
```diff
@@ -27,19 +27,40 @@ const settings = {
...
+processor.use([typedPlugin, typedSetting])
+processor.use([[typedPlugin, typedSetting], [typedPlugin, typedSetting]])
+processor.use([typedPlugin])
+
+// $ExpectError
+processor.use(typedPlugin, settings)
+// $ExpectError
+processor.use([typedPlugin, settings])
+// $ExpectError
+processor.use([[typedPlugin, settings], [typedPlugin, settings]])
```
ChristianMurphy: re-arranged tests to group plugins with untyped testings and plugins with typed settings next to each other.

types/index.d.ts
Rokt33r: > set the generic to unknown or any and lose the strong typing in this case

I think this is the best.

types/tslint.json
Rokt33r: I'm in. This option is kind of helpful but really annoying. After adopting this, all we need to do is more careful when reviewing.

types/unified-tests.ts
Rokt33r: LGTM.

wooorm:
Options get *even harder*. 🤷‍♂️ Will these types support that?

There are plugins that support non-objects as settings. And there are plugins that support multiple arguments

For example,

* [`unified`](https://github.com/unifiedjs/unified/blob/38d2c8211e240a64891aea871caae376428d0200/index.js#L110-L116) — Supports `true` and `false` as well
* [`remark-rehype`](https://github.com/remarkjs/remark-rehype#api) — Supports multiple arguments, the first could be a processor
* [`unified-lint-rule`](https://github.com/remarkjs/remark-lint/blob/1da99f15233b546ddd1552e4ff17da7ec7172bc9/packages/unified-lint-rule/index.js#L69-L100) (and lint rules created with it, such as those in remark-lint) — Supports arrays, numbers, strings

Rokt33r:
@wooorm 
> There are plugins that support non-objects as settings.

This shouldn't be problem.

```ts
interface SomePluginObjectOptions {
  ...
}

type SomePluginOptions = SomePluginObjectOptions | boolean
```

> There are plugins that support multiple arguments.

This might be difficult though... I guess we have two options.

1. Extend unified's type definition from remark-rehype. (See `module-augmentation` http://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation)
2. Add types and documentation all possible usage of unified in here.

--- 

@ChristianMurphy 
While writing this comment, I've noticed that we might be able to use the `module-augmentation` rather than adding generic to use.

With this, the type definition of `use` can be defined in each plugin.

```ts
// *.d.ts in remark-parse
declare module 'unified' {
  interface Processor {
    use(RemarkParseAttacher, RemarkParseOptions): Processor
  }
}
```

I'll prepare some examples tonight so we could review it together.

types/index.d.ts
ChristianMurphy: Updated to use `any`

Rokt33r:
@ChristianMurphy @wooorm 
I prepared the example. As you see, it works quite nicely.
https://github.com/Rokt33r/module-augment-for-unified

But I think using generics with module-augmentation is NOT a good idea.
https://github.com/Rokt33r/module-augment-for-unified/blob/master/node_modules/%40types/unified/index.d.ts#L8

If `use` accepts generics for options, augmented types become completely useless... If generics are provided, it will be prioritized than augmented types. If generics are omitted, the plugin will expect option as `Settings`(`{[key: string]: unknown}`) which is almost same to `unknown`. So augmented types won't work again because they are sub types of `Settings`.

But module-augmentation is not a silver bullet. Because we cannot extend `Pluggable` from external modules, `PluggableList` must become `unknown[]`... And if we use `unknown[]`, `PluginTuple` becomes useless. Because `unknown[]` can be super types for all array/tuple types.

So to consider DX, I think using **module augment without generics** should be better although we cannot type-check for the usage with tuple.

ChristianMurphy:
@Rokt33r can you help me understand the Pros of using [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) in this case?

Conceptually I tend to think about unified plugins as a [Strategy Pattern](https://en.wikipedia.org/wiki/Strategy_pattern) in a [Builder Pattern](https://en.wikipedia.org/wiki/Builder_pattern).
The builder pattern allowing the chaining of `use`, `process`, etc.
The strategy pattern meaning that Unified can `use` anything that implements the interface `Plugin` or `Preset`.

Put another way, at a types level, `Plugin` and `Preset` run inside Unified, rather than augmenting Unified.
Generics fit this pattern pretty well, all plugins will implement the `Plugin` interface, and can optionally define what shape/type the settings are in.

My caution with module augmentation, is that I don't see a way to check that community modules are implementing the `Plugin` or `Preset` interface.
For example, a modified version of @Rokt33r's example repo, where the plugin isn't a function, but is random attributes/properties:

```ts
import unified = require('unified')

interface MyCustomPlugin {
  random: string;
  attributes: boolean;
}

interface MyCustomPluginOptions {
  custom?: 'strict_type'
}

declare module 'unified' {
  interface Processor {
    use(plugin: MyCustomPlugin, options: MyCustomPluginOptions): Processor
  }
}

const customPlugin: MyCustomPlugin = {}

// NOTE: This is valid in TypeScript, but would error when the code is run
unified().use(customPlugin, {
  custom: 'strict_type'
})
```

Rokt33r:
> can you help me understand the Pros of using module augmentation in this case?

I think it will prevent end users from making mistake when adopting plugins because they don't need to provide generics. After they installed `remark-parse`, `Processor` will be expecting the plugin should come along with `RemarkParseOptions`. Even if plugin publishers providing good type definitions, it won't work if end users don't set generics properly.

> My caution with module augmentation, is that I don't see a way to check that community modules are implementing the Plugin or Preset interface.

We still can provide interfaces and types, like `Preset` and `Attacher`, which are helping to build a new plugin. I think ensuring that the module working with unified well is the community's(Plugin publishers) responsibility, not unified itself.

> For example, a modified version of @Rokt33r's example repo, where the plugin isn't a function, but is random attributes/properties:

Yeah, it is too exaggerated... I just wanted to show plugin publishers can easily augment own usage of their plugins.

ChristianMurphy:
> I think it will prevent end users from making mistake when adopting plugins because they don't need to provide generics

I'm still confused here. :confused: 
Can you provide an example where a user would need to provide the generic?

> After they installed `remark-parse`, `Processor` will be expecting the plugin should come along with `RemarkParseOptions`.

Isn't this what these tests demonstrate:
```ts
processor.use(typedPlugin, typedSetting)
processor.use([typedPlugin, typedSetting])

// $ExpectError
processor.use(typedPlugin, wrongTypeSettings)
// $ExpectError
processor.use([typedPlugin, wrongTypeSettings])
```

That the processor will validate the settings match the expected interface?

> And if we go with generics, type checking won't work if end users don't set generics properly.

It would, for example:

```ts
const implicitlyTypedPlugin = (settings?: ExamplePluginSettings) => {} 

processor.use(implicitlyTypedPlugin, typedSetting)
processor.use([implicitlyTypedPlugin, typedSetting])

// $ExpectError
processor.use(implicitlyTypedPlugin, settings)
// $ExpectError
processor.use([implicitlyTypedPlugin, settings])
```

works, with no Generic explicitly set.
TypeScript recognizes that `implicitlyTypedPlugin` implements the interface `Plugin` and that `ExamplePluginSettings` should be used for the generic without needing to be told.
I've added a more complete version of the above example to the test cases.

>  I think ensuring that the module working with unified well is the community's(Plugin publishers) responsibility, not unified's one.

I don't see this as an either or.
Unified can provide typings that help both plugin developers and end users.
And plugin developers can write tests to ensure their plugins work.

> Yeah, it is too exaggerated... I just wanted to show plugin publishers can easily augment own usage of their plugins.

It may be over exaggerated, but it demonstrates a real problem.
It encourages plugin developers to overload Unified with typings that may or may not match what Unified can actually do.

It is also more verbose

```ts
import unified = require('unified')

interface MyCustomPlugin {}

interface MyCustomPluginOptions {
  custom?: 'strict_type'
}

declare module 'unified' {
  interface Processor {
    use(plugin: MyCustomPlugin, options: MyCustomPluginOptions): Processor
  }
}

const customPlugin: MyCustomPlugin = {}

// ===================
// VS
// ===================

interface MyCustomPluginOptions {
  custom?: 'strict_type'
}

const implicitlyTypedPlugin = (settings?: MyCustomPluginOptions) => {} 
```

Rokt33r:
@ChristianMurphy  Hmm.. I changed my mind while I was writing this comment. Let's go with generics now.
I thought module augment would be easier for end users because they don't need to know how `Attacher` or `Preset` looks like.
But, with generics, they still don't need to know about it either. What they need to know is just providing generics to `use()` method...And I don't think it is a big deal as long as we provide documentation properly. So I don't mind if we are not going to adopt module augment.

Anyway, if we are going to use generics for `use()` method, we still need to deal with the case of multiple arguments. So if we still don't want to use module augment for this case, we need to do like the below example.

```ts
use<T1, T2>(plugin: Plugin<T1, T2>, options1: T1, options2: T2)
```

Or should we module augment for this kind of special case?

ChristianMurphy:
> we still need to deal with the case of multiple arguments

@wooorm is it possible for the number of parameters to exceed 2?
Is there any plugin that does this today?

> I guess we should overload like the below example.
> ```ts
> use<T1, T2>(plugin: Plugin<T1, T2>, options1: T1, options2: T2)
> ```

I like this idea. :+1: 
Let's support 2 params for now, and in the case where plugins need more than 2 params, module augmentation may be the answer.

types/index.d.ts
```diff
@@ -169,10 +189,11 @@ declare namespace unified {
    *
    * @this Processor context object is set to the invoked on processor.
    * @param options Configuration
+   * @param extraOptions Secondary configuration
    * @returns Optional.
    */
-  interface Attacher {
-    (this: Processor, options?: unknown): Transformer | void
+  interface Attacher<T = Settings, S = undefined> {
```
ChristianMurphy: Defaulted second param to undefined, since most plugins use a single param.

types/unified-tests.ts
```diff
@@ -27,19 +27,97 @@ const settings = {
...
+processor.use([
+  [implicitlyTypedPlugin, typedSetting],
+  [implicitlyTypedPlugin, typedSetting]
+])
+processor.use([[plugin, settings], [implicitlyTypedPlugin, typedSetting]])
+processor.use([implicitlyTypedPlugin])
+
+// NOTE: settings overrides the generic undefined
+// settings value will be unused but TypeScript will not warn
+processor.use(implicitlyTypedPlugin, typedSetting, settings)
```
ChristianMurphy: One very minor drawback of having two generic params, happens when an implicitly typed plugin with one param, is passed a second param, TypeScript doesn't know it should warn that the value is unused.
:notebook: It won't cause any breakage in code, it just leaves the unused value hanging.

Rokt33r: I think it is fine too.

Rokt33r:
@ChristianMurphy @wooorm 

The changes look good to me. Now the only thing we need to decide is how many argument types should provide.

I think, as long as we don't have any plan to limit the number of options, we should provide like 4 or 5 argument types.

wooorm:
> > we still need to deal with the case of multiple arguments
>
> @wooorm is it possible for the number of parameters to exceed 2?
> Is there any plugin that does this today?
>
> — [@ChristianMurphy](https://github.com/unifiedjs/unified/pull/58#issuecomment-508548833)

Yes, it is possible that there are third and other parameters. On the API side, unified and the engine support it, but it’s not used by the plugins developed *inside* the collective. The second parameter is used, such as by [remark-rehype](https://github.com/remarkjs/remark-rehype#api), and [the engine passes a second parameter](https://github.com/unifiedjs/unified-engine/blob/master/doc/plugins.md#plugins), which is only used by [remark-validate-links](https://github.com/remarkjs/remark-validate-links). I’m not very happy with how second parameter of the engine works, and would like to change it (maybe in [“uniflow”](https://github.com/unifiedjs/governance/pull/22#issue-293216867)?)

Rokt33r:
@wooorm 
> I’m not very happy with how second parameter of the engine works
I agree. One parameter should be enough.

We definitely could do like:

```ts
interface RemarkRehypeOptions extends ToHASTOptions {
  destination: Processor
}
```

So if we really want to modify the current behavior in future, we could merge this pr with two option params for now.
Btw, how should we decide the modification? Should I create an issue first so we could do some poll?

ChristianMurphy:
@wooorm @Rokt33r So I'm hearing a few things:

1. Two parameters should be enough for the use cases we know about, this PR currently supports two params, more than two parameters can be handled by module extension.
2. We should open issue(s) to deprecate and remove usage of the second parameter.
3. After 2. is completed the second parameter can be removed (maybe a major release?).

Is this accurate?
Is this PR ready to be merged?

wooorm:
1 is accurate!

2 and 3 are not what I meant to say [above](https://github.com/unifiedjs/unified/pull/58#issuecomment-509576225). I meant to say that I don‘t like the engine passing a second argument. Because it interferes with plugins accepting other arguments.

I do think this PR is ready to go, and we could talk more about 2 and 3 somewhere else.

wooorm:
How should this be released?

ChristianMurphy:
Existing plugins should continue to work, they now have the option to declare a settings interface.
I'd lean toward marking this as SemVer minor, a feature release.

PR #59, opened 2019-07-11
types: allow processors to declare options/settings

ChristianMurphy:
remark, rehype, etc can be passed settings.
This creates a generic type that allows the language specific instances of `unified` to declare what options are supported.

Based off discussion from https://github.com/remarkjs/remark/pull/383#discussion_r302759676

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -5,11 +5,11 @@ import {VFile, VFileContents, VFileOptions} from 'vfile'
 import vfile = require('vfile')
 
 declare namespace unified {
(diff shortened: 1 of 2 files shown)


ChristianMurphy:
@wooorm this adds a new feature, this should be included in a minor release.

ChristianMurphy:
ideally it would be good for this to go out at the same time as #61

wooorm:
If someone can review 61 tonight I’ll cut 8.2.0 tomorrow!

PR #62, opened 2019-07-13
types: allow any number of settings to be passed to a plugin

### LINKED ISSUE #60: Stop allowing extraSettings
> Yes, it is possible that there are third and other parameters. On the API side, unified and the engine support it, but it’s not used by the plugins developed inside the collective. The second parameter is used, such as by remark-rehype, and the engine passes a second parameter, which is only used by remark-validate-links. I’m not very happy with how second parameter of the engine works, and would like to change it (maybe in “uniflow”?)

@wooorm mentioned in https://github.com/unifiedjs/unified/pull/58#issuecomment-509576225

I think allowing extraSettings is giving too much unnecessary freedom.

For example, remark-rehype is accepting two optional settings, a function for a destination processor and an object for ToHast settings. So remark-rehype should check the first argument's type if it is a function or an object. And we need to provide extra types like the below....

```ts
function remark2rehype(settings?: ToHastSettings)
function remark2rehype(destination?: Processor, settings?: ToHastSettings)
function remark2rehype(destinationOrSettings?: Processor | ToHastSettings, settings?: ToHastSettings)
```

In my opinion, it might cause a little tiny bit of confusing for plugin providers and adopters. Haha

If we don't allow extra settings, *remark-rehype could be coerced to become a bit simpler* like the below example.

```ts
interface RemarkRehypeSettings extends ToHastSettings {
  destination: Processor
}

function remark2rehype(settings?: RemarkRehypeSettings)
```

So, I'm down with this idea. But I also admit that **this is a very trivial issue so we can just ignore.**

ChristianMurphy:
uses [variadic function](https://en.wikipedia.org/wiki/Variadic_function) with [tuple based type checking](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#tuples-in-rest-parameters-and-spread-expressions) to ensure settings and parameters match up.

resolves #60 by removing extra settings in favor of a variadic function.

---

There are some trade offs here, I've documented them to the best of my ability below.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-// TypeScript Version: 3.0
+// TypeScript Version: 3.4
 
 import {Node} from 'unist'
 import {VFile, VFileContents, VFileOptions} from 'vfile'
@@ -23,15 +23,12 @@ declare namespace unified {
      *
      * @param plugin unified plugin
      * @param settings Configuration for plugin
-     * @param extraSettings Additional configuration for plugin
      * @typeParam S Plugin settings
-     * @typeParam S2 Extra plugin settings
      * @returns The processor on which use is invoked
      */
-    use<S = Settings, S2 = undefined>(
-      plugin: Plugin<S, S2, P>,
-      settings?: S,
-      extraSettings?: S2
+    use<S extends any[] = [Settings?]>(
+      plugin: Plugin<S, P>,
+      ...settings: S
     ): Processor<P>
 
     /**
@@ -39,17 +36,16 @@ declare namespace unified {
      *
      * @param preset `Object` with an plugins (set to list), and/or an optional settings object
      */
-    use(preset: Preset<P>): Processor<P>
+    use<S extends any[] = [Settings?]>(preset: Preset<S, P>): Processor<P>
 
     /**
      * Configure using a tuple of plugin and setting(s)
      *
      * @param pluginTuple pairs, plugin and settings in an array
      * @typeParam S Plugin settings
-     * @typeParam S2 Extra plugin settings
      */
-    use<S = Settings, S2 = Settings>(
-      pluginTuple: PluginTuple<S, S2, P>
+    use<S extends any[] = [Settings?]>(
+      pluginTuple: PluginTuple<S, P>
     ): Processor<P>
 
     /**
@@ -214,13 +210,11 @@ declare namespace unified {
    *
    * @this Processor context object is set to the invoked on processor.
    * @param settings Configuration
-   * @param extraSettings Secondary configuration
    * @typeParam S Plugin settings
-   * @typeParam S2 Extra plugin settings
    * @typeParam P Processor settings
    * @returns Optional Transformer.
    */
-  type Plugin<S = Settings, S2 = undefined, P = Settings> = Attacher<S, S2, P>
+  type Plugin<S extends any[] = [Settings?], P = Settings> = Attacher<S, P>
 
   /**
    * Configuration passed to a Plugin or Processor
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// TypeScript Version: 3.0
+// TypeScript Version: 3.4
```
ChristianMurphy: This also works in TypeScript 3.0, but Variadic parameter length checking wasn't added until 3.4

So the test on line 
```ts
// $ExpectError
processor.use(implicitlyTypedPlugin, typedSetting, settings)
```

doesn't get it's expected error

```
ERROR: 102:1  expect  TypeScript@3.1: Expected an error on this line, but found none.
```

this could be reverted back to `3.0` and a note added over the test.

types/index.d.ts
```diff
@@ -253,31 +247,34 @@ declare namespace unified {
...
-  type PluginTuple<S = Settings, S2 = undefined, P = Settings> =
-    | [Plugin<S, undefined, P>, S]
-    | [Plugin<S, S2, P>, S, S2]
+  type PluginTuple<S extends any[] = [Settings?], P = Settings> = [
+    Plugin<S, P>,
+    /**
+     * NOTE: ideally this would be S instead of any[]
+     * As of TypeScript 3.5.2 generic tuples cannot be spread
+     */
+    ...any[]
```
ChristianMurphy: Trying to use `S` here currently results in
```
A rest element type must be an array type
```

This may be a bug, it may be a feature that is not implemented.

---

The trade off here, is we lose variadic tuple type checking of settings, when using tuple syntax (mildly ironic :stuck_out_tongue_winking_eye: ).
This is reflected in the tests where
```ts
processor.use([typedPlugin, invalidSettings])
```

is not caught by TypeScript.

types/index.d.ts
```diff
@@ -253,31 +247,34 @@ declare namespace unified {
...
+    | Preset<S, P>
+    | PluginTuple<S, P>
 
   /**
    * A list of plugins and presets
    *
    * @typeParam P Processor settings
    */
-  type PluggableList<P = Settings> = Array<Pluggable<any, any, P>>
+  type PluggableList<P = Settings> = Array<Pluggable<[any?], P>>
```
ChristianMurphy: a quick worth noting here
this cannot be
```
Array<Pluggable<any[], P>>
```

`any[]` has length `number` (infinity), tuples have a fixed length, which causes inferred `PluggableList`s to fail.
Giving a fixed length `[any?]` works for inference.

types/unified-tests.ts
```diff
@@ -30,7 +30,7 @@ const settings = {
 interface ExamplePluginSettings {
   example: string
 }
-const typedPlugin: Plugin<ExamplePluginSettings> = function() {}
+const typedPlugin: Plugin<[ExamplePluginSettings?]> = function() {}
```
ChristianMurphy: Note that plugin authors, choosing to add an explicit type definition would need to switch to defining settings in a tuple.
Plugin authors leveraging inferencing would see no difference.

types/unified-tests.ts
```diff
@@ -91,16 +94,12 @@ processor.use([pluginWithTwoSettings])
...
-// $ExpectError
 processor.use(typedPlugin, typedSetting, settings)
-// $ExpectError
-processor.use([typedPlugin, typedSetting, settings])
 
 // $ExpectError
 processor.use(implicitlyTypedPlugin, settings)
 // $ExpectError
-processor.use([implicitlyTypedPlugin, settings])
+processor.use(implicitlyTypedPlugin, typedSetting, settings)
```
ChristianMurphy: Implicitly typed plugins now gain the ability to detect when the number of parameters passed is incorrect.

types/index.d.ts
ChristianMurphy: > ```
> A rest element type must be an array type
> ```

There is an issue upstream tracking this https://github.com/microsoft/TypeScript/issues/26113

ChristianMurphy:
This adds a new feature, if merged it should go in a minor release.

PR #64, opened 2019-07-16
Fix Parser and Compiler types

clavin:
The current typings are defined such that [Parsers](https://github.com/unifiedjs/unified#processorparser) and [Compilers](https://github.com/unifiedjs/unified#processorcompiler) (and their respective functional variants) take `VFileCompatible` arguments (and, in the case of Parsers, has the wrong number of arguments). The truth is the spec states that they should accept `VFile` objects (and Parsers take two arguments), and the underlying implementation aligns with this:

* [Parsers are passed a `vfile()`'d argument](https://github.com/unifiedjs/unified/blob/b07233d0c32d47f4acb38cc51ca965e8157230f4/index.js#L270-L274)
* [Compilers are passed a `vfile()`'d argument](https://github.com/unifiedjs/unified/blob/b07233d0c32d47f4acb38cc51ca965e8157230f4/index.js#L340-L344)
* The `file` passed to both is `vfile(doc)` in both.
* Neither type is instantiated/called elsewhere in `index.js`.

Looking for the same inconsistency in the typings, I checked other occurrences of `VFileCompatible` in the `index.d.ts` file. The remaining occurrences align with the spec--that is, anywhere `VFileCompatible` shows up in the typings, the spec explicitly states "any value accepted by `vfile()`".

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -78,7 +78,7 @@ declare namespace unified {
      *
      * `Parser` can also be a constructor function (a function with keys in its `prototype`) in which case it’s invoked with `new`. Instances must have a parse method which is invoked without arguments and must return a `Node`.
      */
-    Parser: ParserFunction | typeof Parser
+    Parser: ParserConstructor | ParserFunction
 
     /**
      * Compile a syntax tree to text.
@@ -98,7 +98,7 @@ declare namespace unified {
      * `Compiler` can also be a constructor function (a function with keys in its `prototype`) in which case it’s invoked with `new`.
      * Instances must have a `compile` method which is invoked without arguments and must return a `string`.
      */
-    Compiler: CompilerFunction | typeof Compiler
+    Compiler: CompilerConstructor | CompilerFunction
 
     /**
      * Transform a syntax tree by applying plugins to it.
@@ -318,20 +318,32 @@ declare namespace unified {
   /**
    * Transform file contents into an AST
    */
-  class Parser {
+  interface Parser {
     /**
      * Transform file contents into an AST
      *
-     * @param text Text to transfomr into AST node(s)
-     * @param file File associated with text
      * @returns Parsed AST node/tree
      */
-    parse(text: string, file: VFile): Node
+    parse(): Node
+  }
+
+  /**
+   * A constructor function (a function with keys in its `prototype`) or class that implements a
+   * `parse` method.
+   */
+  interface ParserConstructor {
+    /**
+     * Creates a Parser
+     *
+     * @param text Text to transform into AST node(s)
+     * @param file File associated with text
+     */
+    new (text: string, file: VFile): Parser
   }
 
   /**
    * Transform file contents into an AST
-   * 
+   *
    * @param text Text to transform into AST node(s)
    * @param file File associated with text
    * @returns Parsed AST node/tree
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -340,21 +345,21 @@ declare namespace unified {
...
      * Transform an AST node/tree into text
      *
-     * @param node Node to be stringified
+     * @param node Node/tree to be stringified
      * @param file File associated with node
-     * @returns transformed text
+     * @returns Compiled text
      */
-    compile(node: Node, file?: VFileCompatible): string
+    compile(node: Node, file: VFile): string
```
ChristianMurphy: do these get passed to the compile method? looking at
https://github.com/unifiedjs/unified/blob/6c84a1f118eaed1606fa17a8d10f2bbef66d7ad3/index.js#L341
it looks like these params are passed to the class initializer rather than the compile method.

types/index.d.ts
```diff
@@ -322,16 +322,21 @@ declare namespace unified {
     /**
      * Transform file contents into an AST
      *
-     * @param file File to transform into AST node(s)
+     * @param text Text to transfomr into AST node(s)
+     * @param file File associated with text
+     * @returns Parsed AST node/tree
      */
-    parse(file: VFileCompatible): Node
+    parse(text: string, file: VFile): Node
```
ChristianMurphy: Do these get passed to the parse method? looking at
https://github.com/unifiedjs/unified/blob/6c84a1f118eaed1606fa17a8d10f2bbef66d7ad3/index.js#L271
it looks like these are passed to the constructor, and that `parse` takes no args.

types/index.d.ts
```diff
@@ -322,16 +322,21 @@ declare namespace unified {
...
   /**
    * Transform file contents into an AST
-   * @param file File to transform into AST node(s)
+   * 
+   * @param text Text to transform into AST node(s)
+   * @param file File associated with text
+   * @returns Parsed AST node/tree
    */
-  type ParserFunction = (file: VFileCompatible) => Node
+  type ParserFunction = (text: string, file: VFile) => Node
```
ChristianMurphy: :+1: 
https://github.com/unifiedjs/unified/blob/6c84a1f118eaed1606fa17a8d10f2bbef66d7ad3/index.js#L274

types/index.d.ts
```diff
@@ -340,21 +345,21 @@ declare namespace unified {
...
    * Transform an AST node/tree into text
    *
-   * @param node Node to be stringified
+   * @param node Node/tree to be stringified
    * @param file File associated with node
-   * @returns transformed text
+   * @returns Compiled text
    */
-  type CompilerFunction = (node: Node, file?: VFileCompatible) => string
+  type CompilerFunction = (node: Node, file: VFile) => string
```
ChristianMurphy: :+1: 
https://github.com/unifiedjs/unified/blob/6c84a1f118eaed1606fa17a8d10f2bbef66d7ad3/index.js#L344

clavin:
Oh! 😳Nice catch on the object methods. Those should be patched up 👍

I also noticed that the `Parser` and `Compiler` types were defined as classes (which they aren't--you cant define a class that `extends` some `Compiler` value defined in unified) so I swapped them over to more appropriate interfaces 😊.

ChristianMurphy:
@wooorm these are fixes and can go in a patch release.

Issue #71, opened 2019-10-01
dtslint silently started failing


wooorm:
### Subject of the issue

dtslint currently fails:

```sh
> unified@8.4.0 test-types unified
> dtslint types

Error: unified/types/index.d.ts:3:1
ERROR: 3:1     strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 4:1     strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 5:8     strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 7:19    strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 412:18  strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
ERROR: 413:1   strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
```

### Your environment

*   **OS**: macOS Mojave 10.14.6
*   **Packages**: latest
*   **Env**: Node 12.2, npm 6.11

### Steps to reproduce

[...]

PR #72, opened 2019-10-01
fix dslint tests

ChristianMurphy:
fix gh-71

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -219,7 +219,7 @@ declare namespace unified {
   /**
    * Configuration passed to a Plugin or Processor
    */
-  type Settings = {
+  interface Settings {
     [key: string]: unknown
   }
 
--- a/types/tslint.json
+++ b/types/tslint.json
@@ -1,7 +1,6 @@
 {
   "extends": "dtslint/dtslint.json",
   "rules": {
-    "callable-types": false,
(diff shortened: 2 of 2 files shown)


types/tslint.json
```diff
@@ -1,15 +1,14 @@
...
     "no-redundant-jsdoc": false,
     "no-void-expression": false,
     "only-arrow-functions": false,
     "semicolon": false,
     "unified-signatures": false,
     "whitespace": false,
-    "interface-over-type-literal": false,
-    "no-unnecessary-generics": false
+    "no-unnecessary-generics": false,
+    "strict-export-declare-modifiers": false
```
ChristianMurphy: this is throwing false positives.
```
export = unified;
```

means that

```
An export assignment cannot be used in a module with other exported elements.
```

types/index.d.ts
```diff
@@ -219,7 +219,7 @@ declare namespace unified {
   /**
    * Configuration passed to a Plugin or Processor
    */
-  type Settings = {
+  interface Settings {
```
ChristianMurphy: this allows `interface-over-type-literal` to be re-enabled

wooorm:
Thanks @ChristianMurphy, @Rokt33r!

PR #73, opened 2019-10-01
ci: add GitHub actions continuous integration workflow

ChristianMurphy:
Since unified is part of the actions beta, created a ci workflow, to try it out.
based off https://github.com/redotjs/redot/pull/17

:notebook: we already have Travis CI, so this would be a redundancy/backup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,25 @@
+name: CI
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+jobs:
+  build-and-test:
+    name: '${{ matrix.platform }}: node.js ${{ matrix.nodejs-version }}'
+    strategy:
+      matrix:
+        platform: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


wooorm:
Thanks for working on this @ChristianMurphy!

My opinion is that two CIs doesn’t help much, but it also doesn’t hurt, and actions is in beta.
For now I’d say use both, maybe in the future we can switch to just GH?

ChristianMurphy:
> For now I’d say use both, maybe in the future we can switch to just GH?

For now let's use both.

A longer term goal, when GH actions becomes stable, we could move to just actions.
It may also be interesting to see if adding the workflow to the [.github](https://github.com/unifiedjs/.github) repo could let us setup a default CI for unified projects.

wooorm:
👍

Is default work flows already supported? We can also use github-tools or so to automatically add and manage files in repo's if not!

ChristianMurphy:
> Is default work flows already supported?

Not sure, would need to try it out.

> We can also use github-tools or so to automatically add and manage files in repo's if not!

:+1:

PR #78, opened 2019-11-04
test: cleanup type tests

ChristianMurphy:


ChristianMurphy:
GitHub actions appears to be having a hiccup/outage, MacOS builds are failing before they start.

Issue #79, opened 2019-11-17
Cannot use the same plugin multiple times in a processor


marko-knoebl:
### Cannot use the same plugin multiple times in a processor

I have a use case where I want to apply the same plugin twice in a processor which doesn't seem to work

### Steps to reproduce

Here's a simple example that uses _rehype-format_ twice. Only the first use seems to succeed:

```js
const unified = require("unified");
const rehypeParse = require("rehype-parse");
const rehypeFormat = require("rehype-format");
const rehypeDocument = require("rehype-document");
const rehypeStringify = require("rehype-stringify");

const processor = unified()
  .use(rehypeParse, { fragment: true })
  .use(rehypeFormat)
  .use(rehypeDocument)
  .use(rehypeFormat)
  .use(rehypeStringify);

const input = "<div><h1>hello world</h1></div>";

const output = processor.processSync(input).toString();
console.log(output);
```

this produces the following output:

```html
<!doctype html>
<html lang="en">
<head>
<meta charset="utf-8">
<meta name="viewport" content="width=device-width, initial-scale=1">
</head>
<body>

<div>
  <h1>hello world</h1>
</div>

</body>
</html>
```

note that the original content was formatted, but the rest was not - the second pass of formatting did not occur

The same behavior occurs with other plugins

### Expected behavior

Allow plugins to be used multiple times

If that's not possible, it would be nice if attempting to use a plugin more than once threw an error

Alternatively, this should at least be documented with possible workarounds (I didn't find documentation on this)

wooorm:
Hi there!

Thanks for the minimal reproduction! In this case, you could of course not use `rehype-format` twice, and only use the latter, for a fix. So, what I’m wondering is, was there something where you couldn’t do that, or was it just that it was unexpected?

The reason for not allowing plugins to be used twice, is to allow reconfiguring:

```js
.use(plugin, {alpha: true, bravo: true})
.use(plugin, {bravo: false, charlie: true})

function plugin(options) {
  console.log(options) // {alpha: true, bravo: false, charlie: true}
}
```

This is useful because [presets](https://github.com/unifiedjs/unified#preset) are often combined in a way where it doesn’t make sense to a) throw errors or b) to run multiple times with different configuration, and c) reconfiguring is the most intuitive.

wooorm:
This could definitely be documented better in the readme here!

marko-knoebl:
Thanks for your reply!

[...]

PR #80, opened 2019-11-23
Add docs on repeated `use` calls targeting the same plugin

marko-knoebl:
This adds the description discussed in #79

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -380,18 +380,17 @@ var unified = require('unified')
 
 unified()
   // Plugin with options:
-  .use(plugin, {})
+  .use(pluginA, {x: true, y: true})
+  // Passing the same plugin again merges configuration (to `{x: true, y: false, z: true}`):
+  .use(pluginA, {y: false, z: true})
   // Plugins:
-  .use([plugin, pluginB])
+  .use([pluginB, pluginC])
   // Two plugins, the second with options:
-  .use([plugin, [pluginB, {}]])
+  .use([pluginD, [pluginE, {}]])
   // Preset with plugins and settings:
-  .use({plugins: [plugin, [pluginB, {}]], settings: {position: false}})
+  .use({plugins: [pluginF, [pluginG, {}]], settings: {position: false}})
   // Settings only:
   .use({settings: {position: false}})
-
-function plugin() {}
-function pluginB() {}
 ```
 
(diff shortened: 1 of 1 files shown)


wooorm:
Sorry for letting this sit! (note we have a strict markdown code style so make sure to run `npm test` locally to make sure your markdown passes).

Your line looks good but I think we should also add an example. What do you think of this, at the end of `use`s example section?

````markdown
Note that when the same plugin is given multiple times, it will be attached
once and options are merged:

```js
var unified = require('unified')

unified()
  .use(plugin, {alpha: true, bravo: true})
  .use(plugin, {bravo: false, charlie: true})
  // …

function plugin(options) {
  console.log(options) // {alpha: true, bravo: false, charlie: true}
}
```
````

**Comment by marko-knoebl** [comment]
I'm finally getting back to this again. Just adapted the line length.

I think the example would be a good addition, but I'm not sure where you're suggesting to put it?

**Comment by wooorm** [comment]
I was thinking right after the current example in that section?

So right before this heading: https://github.com/unifiedjs/unified#processorparsefile

**Comment by marko-knoebl** [comment]
One other suggestion - what would you think of including all examples in one piece of code as below? I think it would not add as much "clutter"

```js
var unified = require('unified')

unified()
  // Plugin with options:
  .use(pluginA, {x: true, y: true})
  // Plugins:
  .use([pluginB, pluginC])
  // Two plugins, the second with options:
  .use([pluginD, [pluginE, {}]])
  // Preset with plugins and settings:
  .use({plugins: [pluginF, [pluginG, {}]], settings: {position: false}})
  // Settings only:
  .use({settings: {position: false}})
  // Updating options of pluginA (to {x: true, y: false, z: true}):
  .use(pluginA, {y: false, z: true})
```

wooorm:
That’s a great idea!! ✨

I would do it immediately after `pluginA` tho:

```js
  // Plugin with options:
  .use(pluginA, {x: true, y: true})
  // Passing the same plugin again merges configuration (to `{x: true, y: false, z: true}`):
  .use(pluginA, {y: false, z: true})
```

Like so maybe?

marko-knoebl:
Sounds good, I've incorporated the changes

wooorm:
Thank you!

PR #82, opened 2019-12-31
Update extend, which has a CVE

jamiek-acl:
Update extend, which has a CVE: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-16492

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   ],
   "dependencies": {
     "bail": "^1.0.0",
-    "extend": "^3.0.0",
+    "extend": "^3.0.2",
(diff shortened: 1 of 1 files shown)


wooorm:
I don’t think this is really a problem, especially for us, see the earlier comments in https://github.com/justmoon/node-extend/pull/48.
Also: see our [security policy here](https://github.com/unifiedjs/.github/blob/master/security.md).
And finally, your fixes doesn’t fix anything. We are using `^` in semver, so anyone doing `npm install` should get the latest “patched” version anyway.

PR #87, opened 2020-03-26
Set `file.result` when processing to non-text

wooorm:
unified is typically, but not always, used to *serialize* when it compiles.  Compilers don’t always return such a serialized (`string`, `Buffer`) though.  The previous behavior was to place the result of the stringify phase when processing on `file.contents`.  This doesn’t make sense for non-text.  This change places non-text results on `file.result`.

Closes vfile/vfile#45.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,11 @@
 'use strict'
 
-var extend = require('extend')
 var bail = require('bail')
-var vfile = require('vfile')
-var trough = require('trough')
+var buffer = require('is-buffer')
+var extend = require('extend')
 var plain = require('is-plain-obj')
+var trough = require('trough')
+var vfile = require('vfile')
 
 // Expose a frozen processor.
 module.exports = unified().freeze()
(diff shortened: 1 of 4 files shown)


wooorm:
> > checks if the result is `string`, `Buffer`, or nully, and based on that either sets `vfile.contents` or `vfile.result`
> 
> This may work.
> Thinking of edge cases
> 
> 1. what if the return type is a class instance which implements a custom `toString`?
> 
> ```js
> class Example {
>   toString() {
>     return "<h1>hello world</h1>";
>   }
> }
> 
> // where the compiler returns
> new Example();
> 
> // which when used as a string
> `${new Example()}`
> // logs "<h1>hello world</h1>"
> ```
> 
> Is this something that would be considered a `value` or a `result`?
> 
> 2. Depending on how you interpret "nully", would that include falsy values? Like `number` with `0`, `boolean` with `false`, would those be split? (E.G. `0` would go to `vfile.contents`, `1` to `vfile.result`)
>
> — [@ChristianMurphy](https://github.com/vfile/vfile/issues/45#issuecomment-604485649)

---

> 1.

Not handled differently. It’s a non-text (neither string nor buffer), so it goes into `result`. So, if some vdom node was returned that has a `toString` field or function in its `__proto__`, those will go into `result`.

> 2.

Nully values are ignored (so a compiler could signal that the current `file.value` or `file.result` are not to be modified), string and buffer are new values for `file.contents`, and other values all go into `file.result`.

PR #89, opened 2020-06-09
Fixing the broken learn link

trebeljahr:
I think the link to the learning page is broken at the moment because of the black lives matter page which does not have the [learn] URL of https://unifiedjs.com/learn so for the time being a link to the GitHub section or something similar could be added as a kind of fallback.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -25,7 +25,7 @@ Some notable users are [Node.js][], [ZEIT][], [Netlify][], [GitHub][],
 
 *   To read about what we are up to, follow us [Twitter][]
 *   For a less technical and more practical introduction to unified, visit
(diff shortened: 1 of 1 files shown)


wooorm:
Activism is annoying to the status quo; I believe directing traffic away from the website / that activism is counter productive.

Our website is useful but not essential. I propose waiting a bit longer before reverting, optionally with a banner, e.g., as long as [Node’s homepage](https://nodejs.org/en/) is different, too

wooorm:
Going to close this, as this link won’t have to be changed. Open to discussing when and how to change the website somewhere else though!

Thanks @trebeljahr for working on a fix, and for understanding!

PR #92, opened 2020-06-27
types: preset and pluggable list settings type checking

ChristianMurphy:
builds on #91 

allow pluggable list and preset to check settings on first 20 elements

leverages variadic typing and generics to allow the first 20 plugins to have full settings type checking.
plugins after the first twenty can still be added, but no longer have their settings checked.

BREAKING CHANGE: changes to order of the processor settings generic `P`, to go first and plugin settings `S` to go after. Plugins directly using the `Plugin` or `Attacher` type will need to add the first generic with the processor settings.
NOTE: If the plugin does not directly use the `Plugin` or `Attacher` type it may not need to migrate. For example
```ts
// this does NOT need to migrate, it will continue to work
const plugin = () => (settings: Settings) => { /* ... */ }

// this DOES need to migrate, settings would passed to the wrong generic parameter
const plugin: Plugin<[Settings?]> = () => (settings: Settings) => { /* ... */ }
```
MIGRATION: if `this` is not used in the plugin, set the first generic to `unknown`
For example
```typescript
Plugin<[PluginSettings?]>
// becomes
Plugin<unknown, [PluginSettings?]>

// and

Attacher<[PluginSettings?]>
// becomes
Attacher<unknown, [PluginSettings?]>
```
if it does depend on `this`, add the processor options in the first generic, for example `RemarkOptions` or `RehypeOptions` could be used.

:warning: TypeScript 4 is currently in beta, this should not be merged until after there is a stable release.
:warning: Adopting this change would require migration of some plugins' typing

---

Additional Discussion:
1. Is there a way to support an artibrary number of plugins in `PluggableList`? Pointers on the syntax for that would be welcome!
2. Should `Attacher` and `Plugin` keep their current generic order? This would reduce the need for migration, but would introduce incosistency between generic parameter order for `Attacher` and  `Plugin` vs `PluggableList` and `Preset`

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -27,7 +27,7 @@ declare namespace unified {
      * @returns The processor on which use is invoked
      */
     use<S extends any[] = [Settings?]>(
-      plugin: Plugin<S, P>,
+      plugin: Plugin<P, S>,
       ...settings: S
     ): Processor<P>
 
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -266,17 +395,105 @@ declare namespace unified {
...
+    Pluggable<P, S12>?,
+    Pluggable<P, S13>?,
+    Pluggable<P, S14>?,
+    Pluggable<P, S15>?,
+    Pluggable<P, S16>?,
+    Pluggable<P, S17>?,
+    Pluggable<P, S18>?,
+    Pluggable<P, S19>?,
+    Pluggable<P, S20>?,
+    ...Array<Pluggable<P, any[]>>
```
ChristianMurphy, line 495: allows more plugins to be added past 20, but without enhanced settings type checking

PR #93, opened 2020-07-20
Add `FrozenProcessor` interface to types

remcohaszing:
A frozen processor has the same interface as a regular processor, except `use()`
is always disallowed.

This is related to remarkjs/remark#513

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -57,7 +57,7 @@ declare namespace unified {
   }
 
   /**
-   * A frozen processor is just like a regular processor, except no additional plugins can’t be added.
+   * A frozen processor is just like a regular processor, except no additional plugins can be added.
    * A frozen processor can be created by calling `.freeze()` on a processor.
    *
    * @see Processor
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -62,6 +54,22 @@ declare namespace unified {
      * @param processorSettings Settings passed to processor
      */
     use(processorSettings: ProcessorSettings<P>): Processor<P>
+  }
+
+  /**
+   * A frozen processor is just like a regular processor, except no additional plugins can’t be added.
```
wooorm, line 60: Typo: no...can't

remcohaszing:
One note by the way. `.use()` **does** exist on frozen processors. It just always throws. Because of this I think this might be a good change, but I’m not 100% sure.

ChristianMurphy:
> does exist on frozen processors. It just always throws

I think removing it from the typing is inline with the spirit of TypeScript.
For example, how TS handles objects

```ts
const a = {}

a.test
//^^^^
// Property 'test' does not exist on type '{}'.
```

`.test` throws a compile time error, even though one could access `a.test` and always get `undefined`.

remcohaszing:
> we _could_ add `use` with return type `never` in the `FrozenProcessor` typing https://mariusschulz.com/blog/the-never-type-in-typescript
> I'm not sure it would add much benefit over not including `use` at all.

I don’t think this adds much benefit.

wooorm:
Released in 9.1.0!

PR #95, opened 2020-07-20
Processor parse type

remcohaszing:
This allows to link libraries based on unified to their matching types.

For example this can be used to declare `remark.parse()` returns mdast and `rehype.parse()` returns hast.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -57,7 +57,7 @@ declare namespace unified {
   }
 
   /**
-   * A frozen processor is just like a regular processor, except no additional plugins can’t be added.
+   * A frozen processor is just like a regular processor, except no additional plugins can be added.
    * A frozen processor can be created by calling `.freeze()` on a processor.
    *
    * @see Processor
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -10,7 +10,7 @@ declare namespace unified {
    *
    * @typeParam P Processor settings. Useful when packaging unified with a preset parser and compiler.
    */
-  interface Processor<P = Settings> extends FrozenProcessor<P> {
+  interface Processor<P = Settings, N = Node> extends FrozenProcessor<P, N> {
     /**
      * Configure the processor to use a plugin and optionally configure that plugin with options.
      *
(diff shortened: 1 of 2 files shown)


remcohaszing:
I added `Node` as default parameter, so the change is non-breaking.

remcohaszing:
I noticed the following outputs mdast:

```js
const remark = require('remark');
const remark2rehype = require('remark-rehype');

const mdast = remark().use(remark2rehype).parse('# Hello');

console.dir(mdast, { depth: Infinity });

// {
//   type: 'root',
//   children: [
//     {
//       type: 'heading',
//       depth: 1,
//       children: [
//         {
//           type: 'text',
//           value: 'Hello',
//           position: Position {
//             start: { line: 1, column: 3, offset: 2 },
//             end: { line: 1, column: 8, offset: 7 },
//             indent: []
//           }
//         }
//       ],
//       position: Position {
//         start: { line: 1, column: 1, offset: 0 },
//         end: { line: 1, column: 8, offset: 7 },
//         indent: []
//       }
//     }
//   ],
//   position: {
//     start: { line: 1, column: 1, offset: 0 },
//     end: { line: 1, column: 8, offset: 7 }
//   }
// }
```

This made me think the parser always outputs the same, but not the compiler or plugins. Of course the fact that `remark-rehype` doesn’t change the parser, doesn’t mean it can’t be done.

I think it should be possible to type plugins and compilers more strictly as well, but that’s more advanced than just the parser.

ChristianMurphy:
> I noticed the following outputs mdast:
> ```
> const mdast = remark().use(remark2rehype).parse('# Hello');
> ```

Correct, `parse` and `stringify` skip the transform step, meaning `remark2rehype` isn't applied.

```
| ........................ process ........................... |
| .......... parse ... | ... run ... | ... stringify ..........|

          +--------+                     +----------+
Input ->- | Parser | ->- Syntax Tree ->- | Compiler | ->- Output
          +--------+          |          +----------+
                              X
                              |
                       +--------------+
                       | Transformers |
                       +--------------+
```

https://github.com/unifiedjs/unified#description

ChristianMurphy:
> This made me think the parser always outputs the same

A parser will always output the same, but it can be replaced via plugin for example

```ts
const remark = require("remark")
const rehypeParse = require('rehype-parse')

const tree = remark()
  .use(rehypeParse)
  .parse("<h1>test</h1>")

console.log(tree)
```

will parse `hast`.

---

and it can be replaced multiple times, for example:

```ts
const remark = require("remark")
const remarkParse = require("remark-parse")
const rehypeParse = require('rehype-parse')

const tree = remark()
  .use(rehypeParse)
  .use(remarkParse)
  .parse("# parser replaced multiple times")

console.log(tree)
```

will be parsed as `mdast`

remcohaszing:
I do think it should be possible to add proper types for this, but this is not the right solution (yet). It’s more complex than this. I don’t think I’ll finalize this anytime soon though. Should I close this PR?

ChristianMurphy:
> I do think it should be possible to add proper types for this

I agree, I think it's possible.

> I don’t think I’ll finalize this anytime soon though. Should I close this PR?

If you plan on re-using many of the changes in this PR, I'd be okay keeping it open.
If it's going to be a PR from scratch, it would be easier to close this and open a fresh PR.

remcohaszing:
I'll close this for now then

PR #96, opened 2020-08-21
WIP: types: use tail recursive variadic tuple to type check pluggable arrays

ChristianMurphy:
alternative to #92 

This approaches the typing from the perspective of tail recursion.
Each call to `PluggableList` will handle the type of the first/head item of the list, the it will recurse for all other items in the list.

TODO:
- [ ] either the syntax is off, or the inference is not currently working, more investigation is needed, help is welcome!

### DIFF
--- a/types/ts4.0/index.d.ts
+++ b/types/ts4.0/index.d.ts
@@ -44,8 +44,15 @@ declare namespace unified {
      * A list of plugins and presets to be applied to processor
      *
      * @param list List of plugins, presets, and pairs
+     * @typeParam S first Plugin's settings
+     * @typeParam L tail recursive list of other plugins
      */
-    use(list: PluggableList<P>): Processor<P>
+    use<
+      S extends any[] = [Settings?],
+      L extends PluggableList<any[], P> = []
+    >(
+      list: PluggableList<S, P, L>
+    ): Processor<P>
 
     /**
      * Configuration passed to a frozen processor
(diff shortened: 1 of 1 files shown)


Murderlon:
This looks cool.

But I’m not sure if I’m up to speed the implementation details. Are these assumptions correct?

with #92 
```ts
// ..
.use<A, B, C>([pluginA, pluginB, pluginC]) // Matches types in the same order. Repetition required, even if types are the same.
// ..
```

with #96 (this)
```ts
// ..
.use<A, R>([pluginA, pluginB, pluginC]) // `A` is matched with pluginA and `R` with pluginB and pluginC
// ..
```

ChristianMurphy:
Excellent question

In both cases the goal is to support type inferencing.
```ts
.use([pluginA, [pluginB, settingsB], presetC])
```
should work.

---

Explicitly showing the types for #92 
```ts
.use<SettingsA, SettingsB, SettingsC>([pluginA, [pluginB, settingsB], presetC])
```

with the limitation that it would no longer check after 10 items are added to `use`.

---

Explicitly showing the types for #96 
```ts
.use<SettingsA, PluggableList<SettingsB, PluggableList<SettingsC, []>>>([pluginA, [pluginB, settingsB], presetC])
```
where PluggableList is a recursive type, which types a single plugin setting, and recurses to support any other settings.
The advantage being that it should support any number of parameters passed.

wooorm:
What is blocking this?

ChristianMurphy:
getting the typing to work 🙂 
closing this for now, will revisit later if there is interest

PR #105, opened 2020-11-04
Remove outdated description of compile results

inokawa:
Currently `rehype-react` uses `vfile.result` instead of `vfile.contents`, and there are same sentances just below.
So I guess these lines are outdated...


wooorm (approved):
whoops, yes that should’ve been removed!

wooorm:
Thanks @inokawa!

Issue #55, opened 2019-06-29
The dependency "is-plane-obj" is not browser friendly


jasonqsong:
### The dependency "is-plain-obj" npm package is not compatible with ...... IE

tl;dr [is-plain-object](https://www.npmjs.com/package/is-plain-object) seems like a good alternative.

Unlike other packages, "is-plain-obj" is publishing arrow functions without transpiling.  It will cause the applications that use IE as the javascript engine failing to load unifiedjs. For my case, it's on-premise MS Word 2019. In addition, although "is-plain-obj" is a 4 line function, it's claiming to [support node8+ only](https://github.com/sindresorhus/is-plain-obj/blob/06c90ec6ec09683bccdbb573d7ed082198297cdb/package.json#L13). I believe unifiedjs is aiming to support both browser and backends, so a backend only package might not be a good dependency.

FYI, my use case is writing a text processor as an Add-on for MS Word with typescript and unified. My current workaround is using babel to load this package first in my webpack.config.js. Which is obviously an anti-pattern because I'm compiling the package in node_modules

```javascript
      rules: [
        {
          test: /node_modules\/.*is\-plain\-obj.*\.js$/,
          use: {
            loader: "babel-loader",
            options: { presets: ["@babel/preset-env"] },
          },
        },
``` 

### Your environment

*   **OS**: Windows 10
*   **Packages**: unified, webpack
*   **Env**: IE11

### Steps to reproduce

1. Create a minimal project

```bash
#!/bin/bash
mkdir -p minimal-project/src
cd minimal-project
echo '<script src="main.js"></script>' > index.html
echo 'import unified from "unified"' > src/index.js
yarn add unified webpack webpack-cli webpack-dev-server
./node_modules/.bin/webpack-dev-server --mode development
```

2. Open Intenet Explorer and navigate to http://localhost:8080

Then see syntax error from is-plain-obj

![image](https://user-images.githubusercontent.com/6178947/60378522-347d4280-9a56-11e9-90b6-7b007b840a45.png)

### Expected behaviour

unifiedjs should load correctly

### Actual behaviour

The 'is-plain-obj' npm package contains ES6 syntax

ChristianMurphy:
Unified can be run in the browser, to do so it needs to be bundled (using webpack, rollup, etc) and transpiled (babel, buble, sucrose, etc) (see https://github.com/unifiedjs/unified/issues/34#issuecomment-364196553).

[...]

Issue #121, opened 2021-02-15
Next major for the ecosystem


wooorm:
Soon (April), it’s time to switch to ESM (without CJS backup) for all packages in the ecosystem.
That’s a lot of projects switching over, and a round of majors that bubbles from the bottom through to the top.

I’ve had some experience with that recently in a push in micromark (with CJS fallback, causing [issues](https://github.com/remarkjs/react-markdown/issues/518)). My recent own projects [dioscuri](https://github.com/wooorm/dioscuri) and [xdm](https://github.com/wooorm/xdm) are fully ESM. It’s going to be a lot of fun for the ecosystem 😅

Otherwise, here are some changes slate for this next bubbling:

**Big breaking changes**:

* archive and deprecate remark plugins that (should) have rehype equivalents, and add some docs somewhere on whether folks should use remark or rehype plugins (currently quite confusing)
* `vfile-message`: rename `location` -> `position` (https://github.com/vfile/vfile-message/issues/6)
* `vfile`: rename `contents` to `value` (https://github.com/vfile/vfile/issues/49)
* `mdast-util-to-string`: do not use `node.title` (https://github.com/syntax-tree/mdast-util-to-string/issues/8)
* `mdast`: rename `depth` to `rank` (https://github.com/syntax-tree/mdast/issues/31) (I feel the other issues there plus https://github.com/syntax-tree/nlcst/issues/6 are too much break for what they bring)
* `hast`: change doctypes (https://github.com/syntax-tree/hast/issues/19)

**Small breaking changes**:

* hastscript’s h and s and a JSX pragma https://github.com/syntax-tree/hastscript/pull/15
* `unist-util-find-all-before`’s results should be reversed probably https://github.com/syntax-tree/unist-util-find-all-before/issues/6

For ES++ features (https://github.com/unifiedjs/rfcs/pull/4), I’d say we can do that after we switch to ESM, because ESM will already be a lot of work, and it gives us a baseline of what ES features / engines to support.

chrisrzhou:
Are there discussion threads where I can read more into the motivation to move formally to using ESM? Inquiring to stay up to date with the unified ecosystem, and also getting context in the broader JS ecosystem.

wooorm:
https://github.com/micromark/micromark/pull/27 and other PRs there, but that was both ESM and CJS.
See also https://twitter.com/sindresorhus/status/1349294527350149121.
The motivation is essentially: we’ll have to move some time. This is a good time!

[...]

PR #134, opened 2021-04-28
#55: swapped is-plain-obj for is-plain-object

mummybot:
Closes GH-55.

I have read the [comment thread](https://github.com/unifiedjs/unified/issues/55#issuecomment-506920290) and seen [the open PR](https://github.com/unifiedjs/unifiedjs.github.io/pull/9) regarding the docs for transpiling individual node_modules. We have it working on our specific project with this technique.

However we have created our own internal library which is shared in a large organisation across multiple teams, all of whom need to support IE11. Because our internal library includes remark, unified and thus **is-plain-obj**, every team will have to make this change in their build tooling to transpile `node_modules/is-plain-obj`.

We would very much appreciate it if you were to swap out the libraries, as it is the simplest of changes and makes unified work in IE11 straight out of the box.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
 var bail = require('bail')
 var buffer = require('is-buffer')
 var extend = require('extend')
-var plain = require('is-plain-obj')
+var {isPlainObject} = require('is-plain-object')
 var trough = require('trough')
 var vfile = require('vfile')
 
@@ -229,7 +229,7 @@ function unified() {
       var entry = find(plugin)
 
       if (entry) {
-        if (plain(entry[1]) && plain(value)) {
+        if (isPlainObject(entry[1]) && isPlainObject(value)) {
           value = extend(true, entry[1], value)
         }
 
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "bail": "^1.0.0",
     "extend": "^3.0.0",
     "is-buffer": "^2.0.0",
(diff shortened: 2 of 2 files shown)


ChristianMurphy:
> However we have created our own internal library which is shared in a large organisation across multiple teams, all of whom need to support IE11. Because our internal library includes remark, unified and thus is-plain-obj, every team will have to make this change in their build tooling to transpile node_modules/is-plain-obj.

A build tool is already needed to use unified in the browser, regardless of this dependency. :confused: 

> We would very much appreciate it if you were to swap out the libraries, as it is the simplest of changes and makes unified work in IE11 straight out of the box.

I don't think it make sense to add an IE11 polyfill for everyone, when build tooling which is already needed does it automatically, and only where needed.

---

at a more overarching level, the unified community is moving towards ES6 features, like ES Modules https://github.com/unifiedjs/unified/issues/121 and `let`/`const`/`class` https://github.com/unifiedjs/rfcs/pull/4.
Which are supported in node and in ever green browsers https://kangax.github.io/compat-table/es6

I appreciate that fact that supporting IE11 can be a challenge, it's something I've faced myself.
But Open Source continuing to support a browser which has been long since abandoned isn't sustainable.

Tools like the ones in https://github.com/unifiedjs/unifiedjs.github.io/pull/9 offer a way forward, assistance improving the documentation is welcome, and if you want to create a shareable build configuration for IE11 support that would be welcome as well.

PR #135, opened 2021-04-28
55: upgraded is-plain-obj

### LINKED ISSUE #55: The dependency "is-plane-obj" is not browser friendly

mummybot:
I'll try again @ChristianMurphy :)

@sindresorhus has upgraded is-plain-obj to no longer use `=>` notation, which resolves both of our issues.
https://github.com/sindresorhus/is-plain-obj/blob/main/index.js

Fixes #55

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "bail": "^1.0.0",
     "extend": "^3.0.0",
     "is-buffer": "^2.0.0",
-    "is-plain-obj": "^2.0.0",
+    "is-plain-obj": "^4.0.0",
     "trough": "^1.0.0",
     "vfile": "^4.0.0"
   },


wooorm:
@mummybot That package is now ESM only. (Which as mentioned, we’ll do soon too). It’ll crash every user of unified 😅

wooorm:
I’ll close this now, I appreciate the work, but I’ll get to it soonish when all the lower-level stuff is done too, and do it all in one working swoop.
But it won’t fix your original issue: IE11.

PR #139, opened 2021-06-18
deps: upgrade is-plain-obj

bertho-zero:
Old `is-plain-obj` use an arrow function and is not compatible with old browsers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "bail": "^1.0.0",
     "extend": "^3.0.0",
     "is-buffer": "^2.0.0",
-    "is-plain-obj": "^2.0.0",
+    "is-plain-obj": "^4.0.0",
     "trough": "^1.0.0",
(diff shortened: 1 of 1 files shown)


wooorm:
New one uses ESM, so will completely break everything.

To support older browsers, compile your code. See for example: https://github.com/remarkjs/react-markdown/issues/359.

bertho-zero:
The ESM is compiled by the bundler, not the ES6 syntax.

bertho-zero:
Bundlers replace exports to make a single file (with or without chunks), but it does not affect the ES6 syntax without a transpiler in any way. In addition, ESM is well understood by bundlers natively and allows better optimization of the build.

wooorm:
Indeed, you need to transpile your code.

See our ESM plans here: https://github.com/unifiedjs/unified/issues/121.

PR #145, opened 2021-07-04
types: add support for boolean plugin options

wooorm:
### Initial checklist

### Description of changes

`.use` can be called with `true` or `false` to turn plugins on or off

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -17,7 +17,7 @@ export interface Processor<P = Settings> extends FrozenProcessor<P> {
    * @typeParam S Plugin settings
    * @returns The processor on which use is called
    */
-  use<S extends any[] = [Settings?]>(
+  use<S extends any[] = [(Settings | boolean)?]>(
     plugin: Plugin<S, P>,
     ...settings: S
   ): Processor<P>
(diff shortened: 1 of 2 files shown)


ChristianMurphy (commented):
A couple questions:
1. is this only for plugins? Or can it also apply to compiler, parser, and/or settings options?
2. Is this mostly to support `remark-lint`? Would it make sense to get typings in place there as an alternative?

wooorm:
1. Plugins: https://github.com/unifiedjs/unified/blob/45eb72e7e7bb4745b94677c5f5681ac0c433d704/index.js#L108-L114
2. `unified-engine`, but yes, also `remark-lint`

ChristianMurphy (approved):

types/index.d.ts
```diff
@@ -221,7 +221,10 @@ export interface FrozenProcessor<P = Settings> {
  * @typeParam P Processor settings
  * @returns Optional Transformer.
  */
-export type Plugin<S extends any[] = [Settings?], P = Settings> = Attacher<S, P>
+export type Plugin<
+  S extends any[] = [(Settings | boolean)?],
```
wooorm, line 225: @ChristianMurphy Hmm, this now results in TS thinking that a plugin can receive a boolean instead of `Settings?`.
I tried removing this change here, to no avail.

Do you know how I `(Settings|boolean)?` can be *passed*, but only `Settings?` received? 🤔

wooorm: Ping @ChristianMurphy

ChristianMurphy: For a single plugin, we'd want to/need to remove boolean from here and add it to

https://github.com/unifiedjs/unified/blob/5167c49ce7e02fcd518adae0acda80ccd4e97400/types/index.d.ts#L20-L23

would probably something like

```ts
 use<S extends any[] = [Settings?]>( 
   plugin: Plugin<S, P>, 
   ...settings: S | [boolean?]
 ): Processor<P> 
```

PR #146, opened 2021-07-05
Add explicit dependency on `@types/unist`

GuiltyDolphin:
### Initial checklist

*   [X] I read the support docs 
*   [X] I read the contributing guide 
*   [X] I agree to follow the code of conduct 
*   [X] I searched issues and couldn’t find anything (or linked relevant results below) 
*   [X] If applicable, I’ve added docs and tests

### Description of changes

This PR adds the latest version of `@types/unist` an explicit dependency (as we import from it directly).

Edit: I notice in the description that dependency on `@types/unist` is _suggested_ for projects using typescript - is there a particular downside to depending on it directly here?

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "index.js"
   ],
   "dependencies": {
+    "@types/unist": "^2.0.5",
     "bail": "^2.0.0",
     "extend": "^3.0.0",
     "is-buffer": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "index.js"
   ],
   "dependencies": {
-    "@types/unist": "^2.0.5",
+    "@types/unist": "^2.0.0",
     "bail": "^2.0.0",
     "extend": "^3.0.0",
     "is-buffer": "^2.0.0",


GuiltyDolphin:
cc @wooorm @remcohaszing @ChristianMurphy

Should I bump unified's (patch?) version number too, so the change in typing doesn't immediately affect packages?

GuiltyDolphin:
Btw @ChristianMurphy @wooorm I'm working through the main unified packages to update `@types/unist` to the latest version. Because it is (technically) a breaking change, updating forces us to make sure we aren't expecting types in the old format, and for going forwards.

I've done a few repos (for the most part, the code doesn't need changing, just the dependencies), but I want to check with you both that I'm doing the correct thing before I go ahead with the others.

package.json
```diff
@@ -46,6 +46,7 @@
     "index.js"
   ],
   "dependencies": {
+    "@types/unist": "^2.0.5",
```
ChristianMurphy, line 49: keep the semver range wide to prevent duplicate versions, where possible 
```suggestion
    "@types/unist": "^2.0.0",
```

ChristianMurphy (approved):

ChristianMurphy:
> Btw @ChristianMurphy @wooorm I'm working through the main unified packages to update @types/unist to the latest version. Because it is (technically) a breaking change, updating forces us to make sure we aren't expecting types in the old format, and for going forwards.
>
> I've done a few repos (for the most part, the code doesn't need changing, just the dependencies), but I want to check with you both that I'm doing the correct thing before I go ahead with the others.

Thanks @GuiltyDolphin!
In general it is worth checking, looking for breaking changes.
We try to keep version ranges wide, for example `^2.0.0` rather than `^2.1.3` for example.
Which means not many repos will need their package.json updated, but a few may need their test suite or typings adjusted.

GuiltyDolphin:
Alright! I'll keep checking through then, but without updating version numbers and just make sure that typings are okay for the latest version of `@types/unist`. It looks like we're mostly okay though, not too many packages are making assumptions about extra keys on `Node`. 😄

remcohaszing (approved):

wooorm (approved):

PR #149, opened 2021-07-06
Fix `next` in types of transformer signature

wooorm:
### Initial checklist

### Description of changes

* `next` can also be called w/o parameters, to just continue on
* `next` returns `void` (`return next()` is fine in a transformer)

### DIFF
--- a/types/unified-tests.ts
+++ b/types/unified-tests.ts
@@ -173,6 +173,9 @@ processor.use(
   // $ExpectError
   () => (tree: Node, file: VFile, next: () => {}) => {}
 )
+processor.use(() => (tree: Node, file: VFile, next: () => void) => {
+  next()
+})
 
 processor.use(pluginWithTwoSettings)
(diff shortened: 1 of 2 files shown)


types/unified-tests.ts
```diff
@@ -155,19 +155,19 @@ processor.use(
...
     (
       tree: Node,
       file: VFile,
-      next: (error: Error | null, tree: Node) => {}
+      next: (error: Error | null, tree: Node) => void
     ) => {}
 )
 processor.use(
-  () => (tree: Node, file: VFile, next: (error: Error | null) => {}) => {}
+  () => (tree: Node, file: VFile, next: (error: Error | null) => void) => {}
```
ChristianMurphy, line 170: could we add a test case where next is called without any params to prevent a regression?

wooorm: done!

ChristianMurphy (approved):
Thanks @wooorm!

PR #156, opened 2021-07-26
Add plugin tree type parameter

wooorm:
### Initial checklist

### Description of changes

This allows plugins to define what type of tree they receive and yield in their transformer.
It does not prevent misuse of those plugins (such as a rehype plugin and then a remark plugin). Nor does it handle parser or compiler plugins.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,11 +25,7 @@ type VFileWithOutput<Result> = Result extends Uint8Array // Buffer.
   : VFile
 
 // Get the right most non-void thing.
-type Specific<Fallback, Left = void, Right = void> = Right extends void
-  ? Left extends void
-    ? Fallback
-    : Left
-  : Right
+type Specific<Left = void, Right = void> = Right extends void ? Left : Right
 
 // Create a processor based on the input/output of a plugin.
 type UsePlugin<
@@ -45,8 +41,8 @@ type UsePlugin<
       // defines a parser, so set `ParseTree`.
       Processor<
         Output,
-        Specific<Node, Output, CurrentTree>,
-        Specific<void, Output, CompileTree>,
+        Specific<Output, CurrentTree>,
+        Specific<Output, CompileTree>,
         CompileResult
       >
     : Input extends Node
@@ -65,8 +61,8 @@ type UsePlugin<
   ? // If `Input` is `Node` and `Output` is not a `Node`, then this plugin
     // defines a compiler, so set `CompileTree` and `CompileResult`
     Processor<
-      Specific<void, Input, ParseTree>,
-      Specific<Node, Input, CurrentTree>,
+      Specific<Input, ParseTree>,
+      Specific<Input, CurrentTree>,
       Input,
       Output
     >
--- a/index.d.ts
+++ b/index.d.ts
@@ -18,39 +18,58 @@ import {VFile, VFileCompatible} from 'vfile'
 
 /* eslint-disable @typescript-eslint/ban-types */
 
-type VFileWithOutput<Result> = Result extends string
+type VFileWithOutput<Result> = Result extends Uint8Array // Buffer.
   ? VFile
-  : Result extends Uint8Array
-  ? VFile
-  : Result extends object
+  : Result extends object // Custom result type
   ? VFile & {result: Result}
   : VFile
 
+// Get the right most non-void thing.
+type Specific<Fallback, Left = void, Right = void> = Right extends void
+  ? Left extends void
+    ? Fallback
+    : Left
+  : Right
+
+// Create a processor based on the input/output of a plugin.
 type UsePlugin<
-  ParseTree extends Node,
-  CurrentTree extends Node,
-  CompileTree extends Node,
-  CompileResult,
-  Input,
-  Output
+  ParseTree extends Node | void = void,
+  CurrentTree extends Node | void = void,
+  CompileTree extends Node | void = void,
+  CompileResult = void,
+  Input = void,
+  Output = void
 > = Output extends Node
   ? Input extends string
     ? // If `Input` is `string` and `Output` is `Node`, then this plugin
       // defines a parser, so set `ParseTree`.
-      // This also assumes that this is the first plugin (so no current tree is
-      // or compile tree is defined yet), so set those too.
-      Processor<Output, Output, Output, CompileResult>
+      Processor<
+        Output,
+        Specific<Node, Output, CurrentTree>,
+        Specific<void, Output, CompileTree>,
+        CompileResult
+      >
     : Input extends Node
     ? // If `Input` is `Node` and `Output` is `Node`, then this plugin defines a
       // transformer, its output defines the input of the next, so set
       // `CurrentTree`.
-      Processor<ParseTree, Output, CompileTree, CompileResult>
+      Processor<
+        Specific<Input, ParseTree>,
+        Output,
+        Specific<CompileTree, Output>,
+        CompileResult
+      >
     : // Else, `Input` is something else and `Output` is `Node`:
       never
   : Input extends Node
   ? // If `Input` is `Node` and `Output` is not a `Node`, then this plugin
     // defines a compiler, so set `CompileTree` and `CompileResult`
-    Processor<ParseTree, CurrentTree, Input, Output>
+    Processor<
+      Specific<void, Input, ParseTree>,
+      Specific<Node, Input, CurrentTree>,
+      Input,
+      Output
+    >
   : // Else, `Input` is not a `Node` and `Output` is not a `Node`.
     // Maybe it’s untyped, or the plugin throws an error (`never`), so lets
     // just keep it as it was.
@@ -61,12 +80,21 @@ type UsePlugin<
 /**
  * Processor allows plugins to be chained together to transform content.
  * The chain of plugins defines how content flows through it.
+ *
+ * @typeParam ParseTree
+ *   The node that the parser yields (and `run` receives).
+ * @typeParam CurrentTree
+ *   The node that the last attached plugin yields.
+ * @typeParam CompileTree
+ *   The node that the compiler receives (and `run` yields).
+ * @typeParam CompileResult
+ *   The thing that the compiler yields.
  */
 export interface Processor<
-  ParseTree extends Node = Node,
-  CurrentTree extends Node = Node,
-  CompileTree extends Node = Node,
-  CompileResult = string
+  ParseTree extends Node | void = void,
+  CurrentTree extends Node | void = void,
+  CompileTree extends Node | void = void,
+  CompileResult = void
 > extends FrozenProcessor<ParseTree, CurrentTree, CompileTree, CompileResult> {
   /**
    * Configure the processor to use a plugin.
@@ -82,7 +110,7 @@ export interface Processor<
    *   *   If the plugin sets a compiler, then this should be the node type that
    *       the compiler expects.
    * @typeParam Output
-   *   Value that is yielded by the plugin.
+   *   Value that the plugin yields.
    *
    *   *   If the plugin returns a transformer, then this should be the node
    *       type that the transformer yields, and defaults to `Input`.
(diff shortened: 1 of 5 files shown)


ChristianMurphy (commented):
:thinking: I like the idea of checking input and output AST types. :+1: 
In an ideal world, unified could do chained type checking, ensuring the output of a transform aligns with the input of the next transform.
parser and compiler plugins will always be a bit funky since their order is independent of transformers, but checking just the transformers could add value. :+1: 

This seems to be a bit different from what is described above though.
It allows `Plugins` to set a generic for their input and output, but this doesn't appear to be checked by unified to ensure the `unified.use` chain is valid?

Is this intended? (from your description it appear so?) If so, could you expand a bit on the value of adding the generics if they are not checked by `unified.use`?

wooorm:
> It allows Plugins to set a generic for their input and output, but this doesn't appear to be checked by unified to ensure the unified.use chain is valid?

Correct.

> Is this intended? (from your description it appear so?) If so, could you expand a bit on the value of adding the generics if they are not checked by unified.use?

So that plugins can start defining what input/output they receive/yield. `unified` otherwise gives/accepts `Node`, which due to:
- index signature removal
- improved mdast/hast (with their registries)
- `hast` / `mdast` utilities now accepting `hast` / `mdast` nodes rather than `Node`
- `unist-util-visit-*` now searching the tree for nodes rather than assuming that they contain a `<Heading>`
…`Node` becomes annoying and too abstract.

---

A future idea I have:
- A `ParserPlugin<OutputTree>` / `CompilerPlugin<InputTree>`:
- Which could perhaps be inferred from `Plugin<[], void, OutputTree>` or `Plugin<[], InputTree, void>` respectively
- or even `Plugin<[], string|Buffer, OutputTree>` or `Plugin<[], InputTree, string|Buffer|unknown>`
…which when `.use`d stores those values on the `Processor<Input, ParseTree, StringifyTree, Output>`, and affects the input/output of `.parse`, `.run`, `.stringify`, and `.process`?

Perhaps this could be includes in this PR.

---

The last step is checking whether the `.use` chains are correct. But because processors are mutable, that’s quite hard:

```
const processor = unified()
processor.use(compilerPlugin)
processor.use(parserPlugin)
processor.use(transformPlugin)
```

Though, most people will probably chain?

ChristianMurphy:
> `unist-util-visit-*` now searching the tree for nodes rather than assuming that they contain a `<Heading>`

Follow up question on this added at https://github.com/syntax-tree/unist-util-visit-parents/pull/10#issuecomment-888310388

> `hast` / `mdast` utilities now accepting `hast` / `mdast` nodes rather than `Node`

A fair point.
I agree this is a challenge, the alternative, adding `is` or `assert` type assertions to every plugin doesn't seem performant or useful.
There is a notable difference between utils and plugins though.
utils are generally directly passed a tree, TypeScript can directly check the types match.
plugins are generally called though `unified`, the tree being indirectly passed, without adding something to unified to try to match up the chained inputs and outputs, we would miss otherwise detectable errors.

> which when `.use`d stores those values on the `Processor<Input, ParseTree, StringifyTree, Output>`, and affects the input/output of `.parse`, `.run`, `.stringify`, and `.process`?

That's what I was thinking as well

> ```ts
> const processor = unified()
> processor.use(compilerPlugin)
> processor.use(parserPlugin)
> processor.use(transformPlugin)
> ```

Each of these `.use()` mutate the `processor`? (I didn't know that was supported)
I'd be fine not supporting changing between node types when using this particular syntax.
I do think it would still be possible to support a single tree/node type with it.

wooorm:
Working on it!

wooorm:
I’ve implemented parsers, compilers, etc for plugins. Now they:
- define the result of `.parse`, `.stringify`, `.process`, `.run`
- define the input of `.run`
- their output defines the input of the next plugin

I have not implemented errors of some sort when misconfiguring plugins.
Given that:
- It is fine to only define transform plugins on a processor because `.run` is used and not `.parse`/`.stringify`
- several parser plugins / several compiler plugins could be used, overwriting what came before
- code is already used a lot in the wild
- the complexity of it all

I’m quite okay with not throwing on `.use(rehypeWhatever).use(remarkWhatever)`

index.d.ts
```diff
@@ -16,16 +16,80 @@
...
+    : // Else, `Input` is something else and `Output` is `Node`:
+      never
+  : Input extends Node
+  ? // If `Input` is `Node` and `Output` is not a `Node`, then this plugin
+    // defines a compiler, so set `CompileTree` and `CompileResult`
+    Processor<ParseTree, CurrentTree, Input, Output>
+  : // Else, `Input` is not a `Node` and `Output` is not a `Node`.
+    // Maybe it’s untyped, or the plugin throws an error (`never`), so lets
+    // just keep it as it was.
+    Processor<ParseTree, CurrentTree, CompileTree, CompileResult>
```
ChristianMurphy, line 57: interesting use of inference to extract generics.
Is this approach used to overcome quirks in JSDoc generics?
Are there other benefits?

index.d.ts
```diff
@@ -82,7 +189,12 @@ export interface Processor extends FrozenProcessor {
  * A frozen processor can be created by calling `.freeze()` on a processor.
  * An unfrozen processor can be created by calling a processor.
  */
-export interface FrozenProcessor {
+export interface FrozenProcessor<
```
ChristianMurphy, line 192: It might be good to document `@typeParam`s for this

index.d.ts
```diff
@@ -16,16 +16,80 @@
...
+    Processor<ParseTree, CurrentTree, CompileTree, CompileResult>
+
+/* eslint-enable @typescript-eslint/ban-types */
+
 /**
  * Processor allows plugins to be chained together to transform content.
  * The chain of plugins defines how content flows through it.
  */
-export interface Processor extends FrozenProcessor {
+export interface Processor<
```
ChristianMurphy, line 65: It might be good to document `@typeParam`s for this

index.d.ts
```diff
@@ -155,9 +270,9 @@ export interface FrozenProcessor {
    *   Nothing.
    */
   run(
-    node: Node,
+    node: ParseTree,
     file: VFileCompatible | undefined,
-    callback: RunCallback
+    callback: RunCallback<CompileTree>
```
ChristianMurphy, line 275: Would this be a `CompileTree`?
`run` doesn't have a compiler?

Or in this context does it mean, final tree type?

index.d.ts
```diff
@@ -382,8 +551,8 @@ export type PluginTuple<PluginParameters extends any[] = any[]> = [
  *   Plugin settings.
  */
 export type Pluggable<PluginParameters extends any[] = any[]> =
-  | PluginTuple<PluginParameters>
-  | Plugin<PluginParameters>
+  | PluginTuple<PluginParameters, any, any>
+  | Plugin<PluginParameters, any, any>
```
ChristianMurphy, line 555: would it make sense/add value to carry the generics through this type as well, rather then `any`s?

ChristianMurphy (commented):
> It is fine to only define transform plugins on a processor

I'm not sure I follow, there is a `CurrentTree` which I assume is the current tree, as output the most recent plugin?
We can assert that it should match the `Input` of the next Plugin?

> several parser plugins / several compiler plugins could be used, overwriting what came before

This impacts parsers and compilers, but transformers should still be fine?

> code is already used a lot in the wild

It is, but these errors would be things which never worked

> the complexity of it all

Most of the complexity seems to come from conditionals to infer, would matching the generics add a lot more complexity?

wooorm:
> We can assert that it should match the Input of the next Plugin?

We could, but existing plugins can be untyped, not typed with input/output, or implicit. I think it’ll cause way to many errors. I’ll check again but what I tried yesterday resulted in 100s of errors in this code base

> It is, but these errors would be things which never worked

Wrong code never worked (rehype plugin before remark plugin without rehype-remark), but I’m not worried about that.
I’m worried about this breaking all the existing plugins.

index.d.ts
```diff
@@ -16,16 +16,80 @@
...
 
+/* eslint-disable @typescript-eslint/ban-types */
+
+type VFileWithOutput<Result> = Result extends string
+  ? VFile
+  : Result extends Uint8Array
+  ? VFile
+  : Result extends object
+  ? VFile & {result: Result}
+  : VFile
```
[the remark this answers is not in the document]
wooorm: Yeah, I think that’s better now.

I had a couple of `VFile & {value: Result}` to set strings/buffers on `file.value`, but I don’t think that’s needed

index.d.ts
wooorm: How else could it be done?

index.d.ts
wooorm: `parse`s output is `ParseTree`
`run`s input is `ParseTree` and output is `CompileTree`
`stringify`s input is `CompileTree` and output is anything.

Open to better wording

index.d.ts
wooorm: yes, I’ll try!

wooorm:
Made the types smarter.
I don’t see how to allow untyped/implicit plugins while also emitting errors when they are chained incorrectly.
I’m not sure how to infer these things from `Pluggable` / `PluggableList` / `Preset` though

index.d.ts
```diff
@@ -171,7 +315,10 @@ export interface FrozenProcessor {
...
    *   Promise that resolves to the resulting tree.
    */
-  run(node: Node, file?: VFileCompatible | undefined): Promise<Node>
+  run(
+    node: Specific<Node, CompileTree>,
+    file?: VFileCompatible | undefined
+  ): Promise<Specific<Node, CompileTree>>
 
   /**
    * Run transforms on the given node, synchroneously.
```
ChristianMurphy, line 324:
```suggestion
   * Run transforms on the given node, synchronously.
```

index.d.ts
```diff
@@ -233,7 +386,7 @@ export interface FrozenProcessor {
    * @returns
    *   Promise that resolves to the resulting `VFile`.
    */
-  process(file: VFileCompatible): Promise<VFile>
+  process(file: VFileCompatible): Promise<VFileWithOutput<CompileResult>>
 
   /**
    * Process a file, synchroneously.
```
ChristianMurphy, line 392:
```suggestion
   * Process a file, synchroneously.
```

index.d.ts
```diff
@@ -560,11 +815,16 @@ export class CompilerClass {
...
+  file: VFile
+) => Result
 
 /**
  * Callback called when a done running.
  *
+ * @typeParam Tree
+ *   The tree that the callback receives.
  * @param error
  *   Error passed when unsuccesful.
```
ChristianMurphy, line 829:
```suggestion
 *   Error passed when unsuccessful.
```

index.d.ts
```diff
@@ -574,25 +834,27 @@ export type CompilerFunction = (tree: Node, file: VFile) => unknown
...
   file?: VFile | undefined
 ) => void
 
 /**
  * Callback called when a done processing.
  *
+ * @typeParam File
+ *   The file that the callback receives.
  * @param error
  *   Error passed when unsuccesful.
```
ChristianMurphy, line 849:
```suggestion
 *   Error passed when unsuccessful.
```

ChristianMurphy (approved):

index.d.ts
ChristianMurphy: `Processor<Input>` each `Plugin<Input>` must match the Processor, `use()` creates a new `Processor<Input>` with the output of the current step in the chain feeding forward.

wooorm: that’s done in:

```ts
export type Plugin<
  PluginParameters extends any[] = any[],
  Input = Node,
  Output = Input
> = (
```

And the `use` overloads:

```ts
use<
  PluginParameters extends any[] = any[],
  Input = Specific<Node, CurrentTree>,
  Output = Input
>
```

…but this conditional makes a *processor*, based on how one is currently configured, and applying those `Input` / `Output` from a plugin

Issue #200, opened 2022-09-24
Throw on preset w/ neither `settings` nor `plugins`


wooorm:
### Initial checklist

- [X] I read the [support docs](https://github.com/unifiedjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/unifiedjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/unifiedjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aunifiedjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Problem

https://github.com/mdx-js/mdx/issues/2136

### Solution

Throw [here](https://github.com/unifiedjs/unified/blob/144eec01b0e1faa23655359d61c3749ad2af99e7/lib/index.js#L200) when neither plugins nor settings is defined.

### Alternatives

Keep as is.

I don’t imagine most folks want empty presets being silently ignored.

For `unified-engine`, we treat config files as presets as well (though importing plugins-as-string). There might be a tiny use case there for empty config files, as they could “restore-to-defaults” a folder, if a higher folder has a config file that prescribes complex things.
This would break that use case.
But I dunno, I don’t think people actually do this?

PR #202, opened 2022-10-12
Add useful error on empty presets

wooorm:
### Initial checklist

### Description of changes

Empty presets are most likely a mistake by the user. At best, they do nothing.
This improves the situation for humans that make mistakes, by throwing a useful error.

Closes GH-200.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -197,6 +197,12 @@ function base() {
      * @returns {void}
      */
     function addPreset(result) {
+      if (!('plugins' in result) && !('settings' in result)) {
+        throw new Error(
+          'Expected usable value but received an empty preset, which is probably a mistake: presets typically come with `plugins` and sometimes with `settings`, but this has neither'
+        )
+      }
+
       addList(result.plugins)
 
       if (result.settings) {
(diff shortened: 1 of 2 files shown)


JounQin (approved):

lib/index.js
```diff
@@ -197,6 +197,12 @@ function base() {
      * @returns {void}
      */
     function addPreset(result) {
+      if (!('plugins' in result) && !('settings' in result)) {
+        throw new Error(
+          'Expected usable value but received an empty preset, which is probably a mistake: presets typically come with `plugins` and sometimes with `settings`, but this has neither'
+        )
+      }
+
```
remcohaszing, line 205: Should we update the types to reflect this?

```ts
interface PresetWithPlugins {
  plugins: PluggableList
  settings?: Record<string, unknown>
}

interface PresetWithSettings {
  plugins?: PluggableList
  settings: Record<string, unknown>
}

export type Preset = PresetWithPlugins | PresetWithSettings
```

remcohaszing:
Since this is semver major, this would also be a good time to considere other breaking changes. The first thing that comes to mind if officially dropping support for Node.js 12.

PR #226, opened 2023-08-15
Test passing a literal object to processor.stringify

remcohaszing:
### Initial checklist

### Description of changes

The object type is correctly inferred as an mdast Root. We should make sure this doesn’t break.

### DIFF
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -575,7 +575,7 @@ const remarkStringify: Plugin<[], MdastRoot, string> = function () {
   // Empty.
 }
 
-expectType<Uint8Array | string>(
+expectType<string>(
   unified().use(remarkStringify).stringify({type: 'root', children: []})
 )
 
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -575,6 +575,10 @@ const remarkStringify: Plugin<[], MdastRoot, string> = function () {
   // Empty.
 }
 
+expectType<Uint8Array | string>(
+  unified().use(remarkStringify).stringify({type: 'root', children: []})
+)
+
 expectType<HastRoot>(unified().use(remarkParse).use(rehypeParse).parse(''))
 
 expectType<string>(


wooorm:
This seems similar to https://github.com/unifiedjs/unified/blob/98ab67534dd4e07fe4032afc1154c1ee8c81dd10/index.test-d.ts#L424-L435?

wooorm:
If you want to test that, we should probably do it in more places. A `const rootLoose = {type: 'root', children: []}` around L32 and it being used in all cases from L327 - L460?

But I’m not sure it needs to be tested. The input of `stringify` is either a specific node such as `MdastRoot` (if a properly typed plugin is used), or a generic unist node (if no properly typed plugin is used). And this object is always assignable to both?

wooorm:
to test assignability of a hast node, you should probably pass a valid root, with an invalid element or so inside it, and check that there’s an error on that element, which wouldn’t happen if it’s a normal unist node?

remcohaszing:
> If you want to test that, we should probably do it in more places

I think to properly test it, all occurrences of `mdastRoot` and `hastRoot` should be replaced with literal. And we should make sure there is no accidental overlap between mdast and hast, so no empty roots.

> But I’m not sure it needs to be tested, here. The input of `stringify` is either a specific node such as `MdastRoot` (if a properly typed plugin is used), or a generic unist node (if no properly typed plugin is used). And this object is always assignable to both?

This tests the type used to properly type a plugin.

It’s true that an `MdastRoot` is assignable to both a generic unist node and `MdastRoot`. That’s exactly the danger. We don’t want to accidentally accept any unist node. `remark().stringify()` should not accept `HastRoot`. It currently doesn’t, but that wasn’t tested.

> And, we have tons of tests in the `test/` folder already

`test/stringify.js` defines an object of type `{ type: string }`, then passes it to `stringify()`. This also tests **assignability**, not type **inference**.

The tests in `test/freeze.js` do, but they only use unist `Node`.

ChristianMurphy:
> I think to properly test it, all occurrences of mdastRoot and hastRoot should be replaced with literal. 

I agree we could and perhaps should test variations of literal.
I'd be cautious removing the tests which use the `mdast` and `hast` node types directly, those are the types which most utilities and plugins will be providing, and it still seems more common to use utilities than fully hand code an AST literal.

wooorm:
> remark().stringify() should not accept HastRoot. It currently doesn’t, but that wasn’t tested.

It’s tested here: https://github.com/unifiedjs/unified/blob/98ab67534dd4e07fe4032afc1154c1ee8c81dd10/index.test-d.ts#L432-L434

> It’s true that an MdastRoot is assignable to both a generic unist node and MdastRoot

It’s also the case that your root object w/o children is both assignable to a generic unist node and a MdastRoot, and a HastRoot

> test/stringify.js defines an object of type { type: string }, then passes it to stringify(). This also tests assignability, not type inference.

Where does type inference come into play here? TS is going to assign in these cases?

remcohaszing:
> > remark().stringify() should not accept HastRoot. It currently doesn’t, but that wasn’t tested.
> 
> It’s tested here:
> 
> https://github.com/unifiedjs/unified/blob/98ab67534dd4e07fe4032afc1154c1ee8c81dd10/index.test-d.ts#L432-L434

I missed this one. This pretty much asserts the same thing and convinces me this PR isn’t needed. So feel free to close.

> > It’s true that an MdastRoot is assignable to both a generic unist node and MdastRoot
> 
> It’s also the case that your root object w/o children is both assignable to a generic unist node and a MdastRoot, and a HastRoot
> 
> > test/stringify.js defines an object of type { type: string }, then passes it to stringify(). This also tests assignability, not type inference.
> 
> Where does type inference come into play here? TS is going to assign in these cases?

This is ok:

```ts
declare function stringify(node: import('mdast').Root): string

stringify({
  type: 'root',
  children: []
})
```

This is not ok:

```ts
declare function stringify(node: import('unist').Node): string

stringify({
  type: 'root',
  children: []  // Type error, the `children` property is excessive
})
```

Also it affects editor completion, hover information, etc.

wooorm:
> This is not ok:
> 
> ```ts
> declare function stringify(node: import('unist').Node): string
> 
> stringify({
>   type: 'root',
>   children: []  // Type error, the `children` property is excessive
> })
> ```

I feel like that’s intentional, a feature not a bug, what we decided on when we removed support for arbitrary fields on `Node` from `unist`.
That’s also why I added a bunch of `NodeLike` and `PositionLike` values to unist utilities.
But now I’m leaning more towards: folks can use types. And some of those utilities should just accept `unknown` and go from their.
And it’s also why we now have `Nodes` in mdast/hast/nlcst, for an enumeration of all valid nodes, instead of the abstract “Node” interface.

remcohaszing:
Yes, the current behaviour a good thing. It was an answer to your question of where the type inference comes into play.

wooorm:
> I missed this one. This pretty much asserts the same thing and convinces me this PR isn’t needed. So feel free to close.

Alright, closing! Thanks for the discussion!

PR #246, opened 2024-06-05
Fix exception on older browsers

### LINKED ISSUE #245: CallableInstance throws error in older ECMA 5 targets
### Initial checklist

- [X] I read the [support docs](https://github.com/unifiedjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/unifiedjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/unifiedjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aunifiedjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

11.0.4

### Link to runnable example

_No response_

### Steps to reproduce

The CallableInstance class is using `Object.defineProperty` without checking `descriptor.configurable`.  When we used unified in a android mobile environment, we saw some differences in how the Processor class was being constructed that caused an error.

The difference we saw was if the javascript is transformed to target ES5 (ECMAScript 5.0), the Processor's `copy` method had a different number of property names defined than in ES6.  Some of these properties are `configurable: false`, so calling `Object.defineProperty` on them throws a "Cannot redefine property: arguments" error.

In ES5:
image

In ES6 (Chrome):
image

Notice in ES6, the number of property names is simply `['length', 'name']`, but in ES5 it includes additional property names that are not configurable.  I believe simply checking `configurable` would solve this issue.

### Expected behavior

Should be able to call `unified()` without error.

### Actual behavior

Calling `unified()` throws an error "Cannot redefine property: arguments" during the `super('copy')` within the Processor's constructor.

### Affected runtime and version

node@20.11.1

### Affected package manager and version

npm@11.0.4

### Affected OS and version

_No response_

### Build and bundle tools

_No response_

justinbhopper:
### Initial checklist

### Description of changes

This addresses errors caused by the unified constructor in ES5 and fixes #245.


wooorm:
Can you add a test?

justinbhopper:
> Can you add a test?

Sure thing -- I've added tests specifically for CallableInstance that should cover this.

justinbhopper:
@wooorm I added some tests, please let me know if this looks good to you

package.json
```diff
@@ -96,7 +96,10 @@
     "atLeast": 100,
     "detail": true,
     "ignoreCatch": true,
-    "strict": true
+    "strict": true,
+    "ignoreFiles": [
+      "test/*.js"
```
ChristianMurphy, line 101: Tests should also be types, it's essential for ensuring the library typing are correct.

test/callable-instance.js
```diff
@@ -0,0 +1,61 @@
+// @ts-nocheck
```
ChristianMurphy, line 1: Similar to above, tests should be type checked.

ChristianMurphy (changes requested):

test/callable-instance.js
justinbhopper: Typescript was insisting that the ES5-style class should be written using ES6 classes, which contradicted the purpose of the test.  I will add @ts-ignore to the specific lines instead.

test/callable-instance.js
```diff
@@ -51,9 +56,12 @@ test('callable-instance', async function (t) {
     const instance = new Es5Class()
     assert.strictEqual(instance.foo, 42)
 
+    /** @type {number} */
     instance.bar = 100
 
     // Instance is callable
+    /** @type {Es5Class} */
+    // @ts-ignore
```
ChristianMurphy, line 64: Could we make this `// @ts-expect-error` with a comment on why it is expected?

justinbhopper: Sure can

ChristianMurphy (approved):

remcohaszing (approved):

wooorm:
Hi again, thanks for your patience!

As mentioned, we don’t support ES5, so it feels weird to include a test for ES5 support.

You said that this problem also occurs more broadly. I wonder if that is the case. Can you come up with a test that shows that to be the case?

In `unified` itself, we don’t use `CallableInstance` on just any value: we only use it on our own `Processor` class, our `copy` function. Adding a `console.log('descriptor:', [p, descriptor?.value])` to the code and running the tests, I get:

```
descriptor: [ 'length', 0 ]
descriptor: [ 'name', 'copy' ]
descriptor: [ 'length', 0 ]
descriptor: [ 'name', 'copy' ]
descriptor: [ 'length', 0 ]
descriptor: [ 'name', 'copy' ]
descriptor: [ 'length', 0 ]
descriptor: [ 'name', 'copy' ]
descriptor: [ 'length', 0 ]
descriptor: [ 'name', 'copy' ]
descriptor: [ 'length', 0 ]
descriptor: [ 'name', 'copy' ]
...
```

Perhaps still your change would make sense to `callable-instance` (and first patched here), but then I worry that by *omitting* non-configurable fields makes your current exception disappear, but will result in more exceptions, because now the fields are omitted.

Or if that isn’t the case, perhaps that we don’t need this descriptor patching loop here at all?

justinbhopper:
@wooorm 

> As mentioned, we don’t support ES5, so it feels weird to include a test for ES5 support.

I added the tests per your request -- are you saying you want me to remove the tests now?

> You said that this problem also occurs more broadly. I wonder if that is the case. Can you come up with a test that shows that to be the case?

Apologies, but I don't recall saying this.  The problem only occurs when `unified` is used in a browser setting and is transpiled to ECMA 5.0.  In fact, I can't even illustrate the bug in a test, because the issue only occurs on browsers and not in Node environment.  

> In unified itself, we don’t use CallableInstance on just any value: we only use it on our own Processor class, our copy function.

Yes, but the bug is specific to CallableInstance.  The Processor class just happens to be the only class inheriting from it.

> Perhaps still your change would make sense to callable-instance (and first patched here), but then I worry that by omitting non-configurable fields makes your current exception disappear, but will result in more exceptions, because now the fields are omitted.

I will make a change so we are instead checking the _target_ object for an existing non-configurable property.  This will ensure we're only skipping fields that are already defined and are non-configurable.  This should ensure we're only skipping non-configurable properties that were defined on a brand new Function.  It won't skip any non-configurable fields from the child class, so we won't be omitting anything from the class itself if they happen to define a non-configurable field.

> Finally: adding your tests but not including your patch to lib, the tests still run for me?

That is because the tests run in Node, and Node does not define Functions with non-configurable properties like browsers do.  I've only encountered this issue in a browser environment (Chrome).  This sandbox illustrates the issue: https://playcode.io/1896334

wooorm:
> I added the tests per your request -- are you saying you want me to remove the tests now?

No, I appreciate a test. But I was wondering if there could be a different test.

> Apologies, but I don't recall saying this. The problem only occurs when unified is used in a browser setting and is transpiled to ECMA 5.0.

See https://github.com/unifiedjs/unified/issues/245#issuecomment-2150635263

> In fact, I can't even illustrate the bug in a test, because the issue only occurs on browsers and not in Node environment.

Right, and this makes me wary: how can we be sure it’s fixed? That this fix does not introduce different problems?

> and Node does not define Functions with non-configurable properties like browsers do.

*You* can still define non-configurable properties in the tests though?

> This sandbox illustrates the issue: playcode.io/1896334

What is the illustration? What behavior do you expect, what not, what is actual? I don’t get errors in Chrome or Safari with that sandbox.

> It won't skip any non-configurable fields from the child class, so we won't be omitting anything from the class itself if they happen to define a non-configurable field.

I worry about this approach: this approach attempts to patch `callable-instance` generically. So, this patch should be possible to upstream. However, this patch just drops some fields without warning. As there is no test, I worry that that would introduce more problems.

I can think of two alternative solutions.

a) Ignore this dead/unneeded code. All the tests should still run, so it’s verified.

```diff
-       const names = Object.getOwnPropertyNames(value)
-
-       for (const p of names) {
-         const descriptor = Object.getOwnPropertyDescriptor(value, p)
-         if (descriptor) Object.defineProperty(apply, p, descriptor)
-       }
+       // Not needed for us in `unified`: we only call this on the `copy`
+       // function,
+       // and we don’t need to add its fields (`length`, `name`)
+       // over.
+       // See also: GH-246.
+       // const names = Object.getOwnPropertyNames(value)
+       //
+       // for (const p of names) {
+       //   const descriptor = Object.getOwnPropertyDescriptor(value, p)
+       //   if (descriptor) Object.defineProperty(apply, p, descriptor)
+       // }
```

b) Create an exception for these legacy and unneeded fields:

```diff
        for (const p of names) {
+         // Special non-configurable legacy function fields (see: GH-246).
+         if (p === 'arguments' || p === 'caller') continue;
+
          const descriptor = Object.getOwnPropertyDescriptor(value, p)
```

(we could also probably add `length` here, which is available on non-Chrome, and also seems useless; then the branch is covered on Node too).

justinbhopper:
> You can still define non-configurable properties in the tests though?

No, the bug is actually caused by attempts to redefine properties on the `apply` object, which is a fresh newly created Function:
image

There isn't any non-configurable fields that _I_ can define that would cause this error.  And now that we are checking the descriptor of the `apply` object's properties, the bug fix should be considered quite safe because it won't skip any non-configurable fields from the class/implementation.

> What is the illustration? What behavior do you expect, what not, what is actual? I don’t get errors in Chrome or Safari with that sandbox.

Apologies, I accidentally saved the fix into the playground.  If you look again using Chrome, you will see the original error.
image

> I can think of two alternative solutions.

Your approach 'A' seems best to me.  I didn't realize unified didn't actually _need_ this cloning behavior.  I agree, it seems unnecessary to attempt to clone properties of a function.  Let me know if you would prefer I change the bug fix approach in this PR.

wooorm:
Thanks so much for thinking with me on this!

> Let me know if you would prefer I change the bug fix approach in this PR.

Yes please!

I prefer approach A as well.
With that approach, I believe all the existing tests are enough, so the new tests can be removed.

justinbhopper:
@wooorm My pleasure!  I've removed the tests and taken your suggested approach.

wooorm (approved):

wooorm:
[released](https://github.com/unifiedjs/unified/releases/tag/11.0.5), thanks!

PR #247, opened 2024-07-16
Fix building with TypeScript 5.5

remcohaszing:
### Initial checklist

### Description of changes

We ran into a regression when building unified with TypeScript 5.5. Splitting the overload into 2 separate overloads fixes it.

This doesn’t require a new release. It just fixes the types if they are built now, the types published to npm are fine.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1049,7 +1049,12 @@ export class Processor extends CallableInstance {
    *
    * @overload
    * @param {Plugin<Parameters, Input, Output>} plugin
-   * @param {...(Parameters | [boolean])} parameters
+   * @param {boolean} parameters
+   * @returns {UsePlugin<ParseTree, HeadTree, TailTree, CompileTree, CompileResult, Input, Output>}
+   *
+   * @overload
+   * @param {Plugin<Parameters, Input, Output>} plugin
+   * @param {...Parameters} parameters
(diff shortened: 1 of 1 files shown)


remcohaszing:
I wasn’t aware unified accepts a boolean instead of options with for plugin though :thinking: 

CI fails, because of type-coverage.

wooorm (approved):

remcohaszing:
Apparently `type-coverage` even fails for a good reason.

wooorm:
> I wasn’t aware unified accepts a boolean instead of options with for plugin though 🤔

```
.use(remarkToc, false)
```

That’s how presets can turn plugins off.

---

Build fails tho

remcohaszing:
Does it also accept `true`, or just `false`?

remcohaszing:
I find this weird TBH. This means plugins can never require options.

Anyway, this is a TypeScript regression. I’ll make a reproduction. Maybe it’s best to pin `typescript@~5.4.0` for now.

wooorm:
> This means plugins can never require options.

They’re called *options*, indeed ;)
But it’s not correct, what you say: plugins can choose to disallow the case where they are not configured by throwing.

wooorm:
Ok, I’ll just ping for now!

This repo already has a script to fix the types that TS generates, so perhaps that’s also possible.

wooorm:
thanks for looking into this, remco!

PR #264, opened 2025-07-24
Update for TS7

sandersn:
### Initial checklist

* [x] I read the support docs 
* [x] I read the contributing guide 
* [x] I agree to follow the code of conduct 
* [x] I searched issues and discussions and couldn’t find anything or linked relevant results below 
* [x] I made sure the docs are up to date
* [x] I included tests (or that’s not needed)

### Description of changes

I am testing Typescript 7's JS support, which I've largely rewritten during the switch to Go. The new code processes tags by converting them into synthetic nodes just after parsing their host node. For unified, that means that `@template` tags can't be shared between multiple `@overload` tags like before.

This PR also has to convert function declarations with `@type` on them to variables initialised with a function expression. TS7 doesn't currently support this feature on function declarations--but both unified and svelte use this feature a lot, so I think we'll need to add it before TS7's JS features are ready. At that point the shuffling of functions won't be needed anymore and this PR will be very small.

In addition, https://github.com/microsoft/typescript-go/pull/1444 makes it possible to have rest parameters in `@overload`, and is needed for this change to work in TS7, so it won't compile with the nightly until that PR is merged. 
Edit: it will be available in the 2025-07-25 nightly.

Because TS7 is quite a way off, I don't know whether you'll want to take this PR. I created it to see how hard it would be to update popular JS code that uses TS for checking.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -538,7 +538,6 @@ export class Processor extends CallableInstance {
    *   processor.data() // => {charlie: 'delta'}
    *   ```
    *
-   * @template {keyof Data} Key
    *
    * @overload
    * @returns {Data}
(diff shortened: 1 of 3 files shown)


sandersn:
I reverted the function->const move because I couldn't get it to compile in TS5 and I've decided to add whole-type annotated functions in TS7, so that change won't be needed.

wooorm:
Hi again!

Right, probably makes sense to have the `@templates` on the overloads! So that also works on TS5? And you do now recommend landing this PR for us?

If you are looking at userland code, you might as well also look at https://github.com/unifiedjs/unified/blob/main/script/fix-types.js 😉 It may be especially interesting because it also deals with rest parameters 🤔

sandersn:
Yes, I'd recommend this change everywhere. Shared template tags is a mistake that I've never fixed until TS7 gave me a chance to break JS backward compatibility.

sandersn:
re fix-types.js: it looks like it had to fix up rest parameters with `| undefined` pre-5.3 and with a postfix `[]` pre-5.5?

It sounds like I need to make sure that TS7 works with this type: `@param {Parameters | [boolean]} rest` for rest parameters that might look like `...rest`. Is that correct? I hope so because I think that is its current state.

PR #267, opened 2025-11-11
Accept non mutable array of plugins in the use function

### LINKED ISSUE #266: Readonly Array for PluggableList
### Initial checklist

- [x] I read the [support docs](https://github.com/unifiedjs/.github/blob/main/support.md)
- [x] I read the [contributing guide](https://github.com/unifiedjs/.github/blob/main/contributing.md)
- [x] I agree to follow the [code of conduct](https://github.com/unifiedjs/.github/blob/main/code-of-conduct.md)
- [x] I searched [issues](https://github.com/search?q=user%3Aunifiedjs&type=issues) and [discussions](https://github.com/orgs/unifiedjs/discussions) and couldn’t find anything (or linked relevant results below)

### Affected package

unified@11.0.5

### Steps to reproduce

I'm using the [React Markdown](https://github.com/remarkjs/react-markdown) library which is using `unified` as a dependency, especially the plugins. The main component can take a list of plugins with a `PluggableList` which is mutable. However I believe this is not meant to be modified but to handover an immutable list of plugins.

```tsx
const REMARK_PLUGINS = [remarkGfm] as const

...
const processor = unified()
    .use(REMARK_PLUGINS)
```

Before opening a bug at ReactMarkdown, I want to check in here:
* is there an alternative which is readonly for PluggableList? (I could not find one)
* is PluggableList actually mostly used as readonly and you would consider making it ReadonlyArray? at least that the `use` function would take it as an immutable array?

If you believe that it should actually stay mutable for some reason, I'll reach out to ReactMarkdown.

### Actual behavior

This makes the Typescript check fails as `REMARK_PLUGINS` is readonly (due to the `as const`) while `use` expects a `PluggableList` which is mutable.

### Expected behavior

Typescript should not choke and `use` should accept a readonly array of plugins.

### Runtime

_No response_

### Package manager

_No response_

### Operating system

_No response_

### Build and bundle tools

_No response_

pcorpet:
This change is compatible with previous code, it makes it more obvious that the library does not modify the array of plugins and only iterates over the array.

Fixes #266

### Initial checklist

* [X] I read the support docs 
* [X] I read the contributing guide 
* [X] I agree to follow the code of conduct 
* [X] I searched issues and discussions and couldn’t find anything or linked relevant results below 
* [x] I made sure the docs are up to date (I don't think they need a change)
* [X] I included tests (or that’s not needed)

### Description of changes

Change the type of the first parameter of `use` to accept not only arrays but also readonly arrays (immutable). As the library does not modify it, we can accept this wider type.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -14,6 +14,7 @@ export type {
   Preset,
   ProcessCallback,
   Processor,
+  ReadonlyPluggableList,
   RunCallback,
   // `Settings` is typed and exposed below.
   TransformCallback,
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -548,6 +548,15 @@ expectType<Processor>(
   })
 )
 
+const PLUGINS = [
+  remarkParse,
+  remarkLint,
+  remarkLintImplicit,
+  remarkRehype,
+  rehypeStringify
+] as const
+expectType<Processor>(unified().use(PLUGINS))
+
 /**
  * Register our setting.
  */
--- a/lib/index.js
+++ b/lib/index.js
@@ -80,6 +80,11 @@
  *   List of plugins and presets.
  */
 
+/**
+ * @typedef {ReadonlyArray<Pluggable>} ReadonlyPluggableList
+ *   Readonly list of plugins and presets.
+ */
+
 // Note: we can’t use `callback` yet as it messes up `this`:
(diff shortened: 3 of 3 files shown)


wooorm:
closing, see https://github.com/unifiedjs/unified/issues/266#issuecomment-3549183204

Issue #263, opened 2025-07-05
tihs package uses non-ESM dependencies in a way that is not compatible with pure ESM environments


NullVoxPopuli:
### Initial checklist

- [x] I read the [support docs](https://github.com/unifiedjs/.github/blob/main/support.md)
- [x] I read the [contributing guide](https://github.com/unifiedjs/.github/blob/main/contributing.md)
- [x] I agree to follow the [code of conduct](https://github.com/unifiedjs/.github/blob/main/code-of-conduct.md)
- [x] I searched [issues](https://github.com/search?q=user%3Aunifiedjs&type=issues) and [discussions](https://github.com/orgs/unifiedjs/discussions) and couldn’t find anything (or linked relevant results below)

### Affected package

unified

### Steps to reproduce

I'm building a library that uses unified. So I can't mandate consumers have specific exceptions for unified using cjs dependencies in the dependency graph.

Try vite 6, or rollup with a project that uses `unified` (or any other ESM-based packager (esm.sh with no pre-build, for example))

Problem comes from here: https://github.com/unifiedjs/unified/blob/45d8fa503157d20a656b3fa00e4b010dc0e50748/lib/index.js#L344

I did see this issue: https://github.com/unifiedjs/unified/issues/153
But, because I'm a library author, not an app dev, the solutions in that issue are not possible.

### Actual behavior

This error:
```
[!] RollupError: ../../node_modules/.pnpm/unified@11.0.5/node_modules/unified/lib/index.js (350:7): "default" is not exported by "../../node_modules/.pnpm/extend@3.0.2/node_modules/extend/index.js", imported by "../../node_modules/.pnpm/unified@11.0.5/node_modules/unified/lib/index.js".
https://rollupjs.org/troubleshooting/#error-name-is-not-exported-by-module
../../node_modules/.pnpm/unified@11.0.5/node_modules/unified/lib/index.js (350:7)
348: 
349: import {bail} from 'bail'
350: import extend from 'extend'
            ^
351: import {ok as assert} from 'devlop'
352: import isPlainObj from 'is-plain-obj'
```

### Expected behavior

either
- `extend` gets a default export
- we drop the `extend` library 🙈 

See: https://rollupjs.org/troubleshooting/#error-name-is-not-exported-by-module

### Runtime

node 24

### Package manager

rollup 4.x

### Operating system

doesn't matter

### Build and bundle tools

Rollup

NullVoxPopuli:
Actually there is a _work around_.

[...]

Issue #153, opened 2021-07-20
Dependencies need to be ESM


benmccann:
### Initial checklist

- [X] I read the [support docs](https://github.com/unifiedjs/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/unifiedjs/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/unifiedjs/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Aunifiedjs&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

all

### Steps to reproduce

Provided in https://github.com/sveltejs/kit/issues/1961

### Expected behavior

All dependencies should be ESM since this package declares `"type": "module"`. The package should work out-of-the-box with SvelteKit and Vite

This can be fixed by distributing an ESM version of `is-buffer` and `extend` or by removing those dependencies from this library since they have minimal usage. I've sent a PR for `is-buffer` https://github.com/feross/is-buffer/pull/43 though I don't know yet whether it will be accepted

This could be added a sub-task to https://github.com/unifiedjs/unified/issues/121

### Actual behavior

Vite encounters CJS dependency of this ESM module and fails with the message `module is not defined` as described in https://github.com/vitejs/vite/issues/4113

### Runtime

Node v14, Other (please specify in steps to reproduce)

### Package manager

npm v7

### OS

Linux

### Build and bundle tools

Vite

ChristianMurphy:
Thanks for reaching out @benmccann!

It's also worth noting the Vite is planning to fix this so it would work by default https://github.com/vitejs/vite/issues/3024#issue-860269204 and there is a work around available today without needing to change packages https://github.com/vitejs/vite/issues/3024#issuecomment-860018367

---

> This can be fixed by distributing an ESM version of `is-buffer` and `extend`

I'm happy to see improvement's in unified's dependencies. :+1: 
Thanks for opening feross/is-buffer#43!
Would you be interested in opening a pull request to `extend` as well?

---

> by removing those dependencies from this library since they have minimal usage

I'm happy to change unified's dependencies, if there are security, performance, or stability benefits. :+1: 

I'm cautious changing dependencies only for the sake of working around a short term bug in a particular tool chain.
[...]

PR #273, opened 2026-05-20
Drop `extend` CJS dependency

fuma-nama:
### Initial checklist

* [x] I read the support docs 
* [x] I read the contributing guide 
* [x] I agree to follow the code of conduct 
* [x] I searched issues and discussions and couldn’t find anything or linked relevant results below 
* [x] I made sure the docs are up to date
* [x] I included tests (or that’s not needed)

### Description of changes

Removed the remaining CJS dependency `extend`, since unified itself is ESM-only, it would be great if all deps are also ESM so it doesn't need bundling to work in browser environments.

The `extend` package is being used as a deep merge utility, while its core functionality is being replaced by `Object.assign()`. I think a proper deep merge utility would do better, but `extend` has some behaviours that may differs from existing deep merge utilities, so I rewrote the original utility with modern ESM, shouldn't cause any breaking changes.

I had to lower the test coverage as some edge cases are not easily reached, let me know if this isn't desirable.

### DIFF
--- a/lib/deep-merge.js
+++ b/lib/deep-merge.js
@@ -1,0 +1,78 @@
+import isPlainObj from 'is-plain-obj'
+
+const own = {}.hasOwnProperty
+
+/**
+ * Deeply merge plain objects and arrays.
+ *
+ * @template {Record<string, unknown> | Array<unknown>} Target
+ * @param {Target} target
+ * @param {unknown} source
+ * @returns {Target}
+ */
+export function deepMerge(target, source) {
+  if (!source || typeof source !== 'object') {
+    return target
+  }
+
+  const values = /** @type {Record<string, unknown>} */ (source)
+
+  for (const key of Object.keys(values)) {
+    const value = getOwn(values, key)
+
+    if (value === undefined || value === target) {
+      continue
+    }
+
+    if (Array.isArray(value)) {
+      const existing = getOwn(target, key)
+      const base = Array.isArray(existing)
+        ? /** @type {Array<unknown>} */ (existing)
+        : []
+
+      setOwn(target, key, deepMerge(base, value))
+    } else if (isPlainObj(value)) {
+      const existing = getOwn(target, key)
+      const base = isPlainObj(existing) ? existing : {}
+
+      setOwn(target, key, deepMerge(base, value))
+    } else {
+      setOwn(target, key, value)
+    }
+  }
+
+  return target
+}
+
+/**
+ * @param {Record<string, unknown> | Array<unknown>} target
+ * @param {string} key
+ * @returns {unknown}
+ */
+function getOwn(target, key) {
+  if (key === '__proto__' && !own.call(target, key)) {
+    return undefined
+  }
+
+  return /** @type {Record<string, unknown>} */ (target)[key]
+}
+
+/**
+ * @param {Record<string, unknown> | Array<unknown>} target
+ * @param {string} key
+ * @param {unknown} value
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -87,7 +85,7 @@
     "prepack": "npm run build && npm run format",
     "test": "npm run build && npm run format && npm run test-coverage",
     "test-api": "node --conditions development test/index.js",
-    "test-coverage": "c8 --100 --check-coverage --reporter lcov npm run test-api"
+    "test-coverage": "c8 --96 --check-coverage --reporter lcov npm run test-api"
```
Murderlon: This is not going to be acceptable for some I think. But from my side it's fine if the code is basically as-is from the `extend` package and we're never going to touch it.

lib/deep-merge.js
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Deeply merge plain objects and arrays.
+ *
+ * @template {Record<string, unknown> | Array<unknown>} Target
+ * @param {Target} target
+ * @param {unknown} source
+ * @returns {Target}
+ */
+export function deepMerge(target, source) {
```
Murderlon, line 37: Did you literally copy the contents of `extend`? I would prefer and trust that more than some AI attempt I think 🤔

Murderlon (commented):
Not up to me to decide whether this goes in. Knowing @wooorm, he really likes consistency. So that usually means using `extend` everywhere or nowhere, not just dropping in one package. Luckily it seems there are [only two](https://github.com/search?q=%2F%22extend%22%2F+path%3Apackage.json+%28org%3Aremarkjs+OR+org%3Arehypejs+OR+org%3Aretextjs+OR+org%3Aunifiedjs+OR+org%3Asyntax-tree+OR+org%3Avfile%29&type=code&query=%28%22extend%22+path%3Apackage.json%29+%28org%3Aremarkjs+OR+org%3Arehypejs+OR+org%3Aretextjs+OR+org%3Aunifiedjs+OR+org%3Asyntax-tree+OR+org%3Avfile%29)

lib/deep-merge.js
fuma-nama: yeah I was writing it side-by-side with the original implementation of `extend`

package.json
fuma-nama: perhaps I can add more unit tests to hit the coverage, will take a look later

fuma-nama:
Oops I didn't search in other repositories, I think that will need another PR, would like to hear some inputs from maintainers before doing that. `extend` doesn't actually merge data structures like Map & Set, there's better alternatives for deep merge utils.

Murderlon:
Taking a step back, wondering if we can use `structuredClone`, [browser support](https://caniuse.com/wf-structured-clone) is good, also in Node.js since 17.0.0 🤔

fuma-nama:
I think some libraries might store functions in the `settings` object, I would prefer to not break anything when possible.

package.json
fuma-nama: added more unit tests, back to 100% coverage now.

wooorm:
Sorry but I don‘t think this is possible.

I tried structuredClone. It passed tests. It worked in several of the 100s of repos. Then in failed in some. Some things put functions on data.
There are lots of people writing their own plugins and doing all kinds of things.

This requires lots and lots of tests to have parity with `extend`.
At that point, put it in a new repo, share it with the world. Maybe then it can be used here.

See GH-263, GH-153

